PR #3, opened 2021-07-26
docs: contributing

cristianmtr:


tadejsv (changes requested):
I see some things missing:
- mention of how to download custom model to use for tests (I see it is in tests script, but would be good to list it in instructions, so people don't loose time looking for it)
- how to handle adding executors to hubble

cristianmtr:
Added a note on pushing to Hub.

The custom model thing is too much depending on your Executor. Just download it in the `__init__`, like the others etc.

tadejsv:
@cristianmtr No, I mean for testing - you have in the test script that `download_model.sh` will be used if it exist, this needs to be mentioned. 

Also, I would not consider downloading in `__init__` the best practice for all models - for large models this results in long startup times, and if using `jinahub+docker` this can happen every time, since containers are ephemeral. In this case telling the user to download first (in the Prerequisites section) and to create a volume is prefered.

README.md
```diff
@@ -102,6 +102,16 @@ pods:
...
+## Contributing
+
+If you want to develop your own Executor, please use the [Executor cookiecutter](https://github.com/jina-ai/cookiecutter-jina-hub) to start with. 
+
+If you are an **external** user, this can then go into your own repository. Please do **not** commit to this repository. This is **only** for internal Jina Engineers. 
+
+If you are a **Jina Engineer**, make sure to:
+- add the new executor to the right subfolder. Check [Types](#types)
+- push your initial version to Jina Hub. Use the guide [here](https://github.com/jina-ai/jina/blob/master/.github/2.0/cookbooks/Hubble.md#2-push-and-pull-cli)
+- add the UUID and secret to the secrets store. Make sure `(folder name) == (manifest alias) == (name in secrets store)`
```
tadejsv, line 114: Where is the secrets store?

cristianmtr:
> @cristianmtr No, I mean for testing - you have in the test script that `download_model.sh` will be used if it exist, this needs to be mentioned.
> 

That was only added for that specific case, which should be refactored, as it will affect maintainability if we add it for each folder.

I'll add a note about it.

> Also, I would not consider downloading in `__init__` the best practice for all models - for large models this results in long startup times, and if using `jinahub+docker` this can happen every time, since containers are ephemeral. In this case telling the user to download first (in the Prerequisites section) and to create a volume is prefered.

I'll add a note about that too

nan-wang (approved):
LGTM👍

tadejsv (approved):
LGTM

PR #15, opened 2021-07-27
feat(torch_transformer): support multi-gpus

numb3r3:


jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
```diff
@@ -88,6 +99,14 @@ def __init__(
         )
         self.model.to(torch.device(device))
 
+        self.is_distill_bert = self.model.config.model_type == 'distilbert'
```
tadejsv: PRs should be, if possible, limited to a single issue. The title of your PR is "support multi-gpus" - I think fixing the encoder to support distilbert should be a separate PR

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
```diff
@@ -153,6 +172,12 @@ def _generate_input_tokens(self, texts):
             truncation=True,
             return_tensors='pt',
         )
+
+        if self.is_distill_bert:
+            # DistilBERT doesn’t have token_type_ids, you don’t need to indicate which token belongs to which segment.
+            # https://huggingface.co/transformers/model_doc/distilbert.html
+            del input_tokens['token_type_ids']
```
tadejsv: As I say above, we should move this to a separate PR, but since it's here: why is this needed? What breaks without it?

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
```diff
@@ -62,22 +62,33 @@ def __init__(
...
+
         if device == 'cuda' and not torch.cuda.is_available():
             self.logger.warning(
-                'You tried to use GPU but torch did not detect your'
+                'You tried to use GPU but torch did not detect your '
                 'GPU correctly. Defaulting to CPU. Check your CUDA installation!'
             )
             device = 'cpu'
 
+        if device == 'cuda':
```
tadejsv, line 73: I am not sure this is the way to go. If we were serious about this, we need to add this to every single pytorch executor. And then we need to think of something similar for tensorflow executors, paddle paddle executors, etc. Another issue that I have with this is that someone might want to have multiple smaller models on a single GPU - seems like this right now is doable by specifying `cuda:0` - but such strong difference between `cuda` and `cuda:0` is not what pytorch users would expect.

To me it feels like this is something that should be handled on a lower (or higher, as you prefer) level - in jina core. Something along the lines of a possibility of passing different init arguments to each replica? I would like to hear someone elses' opinion on this

tadejsv (changes requested):

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
```diff
@@ -88,6 +99,12 @@ def __init__(
         )
         self.model.to(torch.device(device))
 
+        # Do warmup round of inference as the first pass is very slow
+        with torch.no_grad():
```
tadejsv, line 103: I never saw that a warmup would be needed. Would you mind creating an example demonstrating this (a colab notebook perhaps)?

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
numb3r3: thanks for your suggestion. Anyway, i found this is a false alarm (the related issue #16 has been closed). I have reverted this change.

tadejsv: All good then!

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
numb3r3: You can observe the necessary of warmup from the benchmark script I created:
```
import time
import click
from jina.logging.profile import used_memory, get_readable_size
from transformers import AutoModel, AutoTokenizer


@click.command()
@click.option(
    "--batch_size",
    default=32,
)
@click.option(
    "--num_threads",
    default=8,
)
@click.option(
    "--device",
    default='cpu',
)
@click.option(
    "--tasks",
    default=10,
)
def main(batch_size, num_threads, device, tasks):
    import torch

    if device == 'cuda' and not torch.cuda.is_available():
        print(
            'You tried to use GPU but torch did not detect your'
            'GPU correctly. Defaulting to CPU. Check your CUDA installation!'
        )
        device = 'cpu'
    print(f'using device: {device}')

    torch.set_num_threads(num_threads)
    tokenizer = AutoTokenizer.from_pretrained("hfl/chinese-roberta-wwm-ext")
    model = AutoModel.from_pretrained("xcjthu/Lawformer")
    model.to(torch.device(device))
    total = 0
    for _ in range(tasks):
        start_t = time.perf_counter()
        start_mem = used_memory(unit=1)

        batch_date = batch_size * ["任某提起诉讼，请求判令解除婚姻关系并对夫妻共同财产进行分割。"]
        inputs = tokenizer(batch_date, return_tensors="pt")
        outputs = model(**inputs)

        elapsed = time.perf_counter() - start_t
        end_mem = used_memory(unit=1)

        level_prefix = ""
        mem_status = f"memory Δ {get_readable_size(end_mem - start_mem)} {get_readable_size(start_mem)} -> {get_readable_size(end_mem)}"
        time_per_sent = elapsed / batch_size
        total += time_per_sent
        print(
            f"[{_}]-th Encoding Time: {elapsed} s ({time_per_sent} s) {mem_status}"
        )
    print(f"==> AVG use {total / tasks} s per sentence")


if __name__ == "__main__":
    main()
```

The command `python benchmark.py --batch_size 8 --num_threads 8` yield the following result:
```
[0]-th Encoding Time: 2.2921026149997488s (0.2865128268749686s per sentence) memory Δ 7.7 MB 1.1 MB -> 8.8 MB
[1]-th Encoding Time: 2.2730195039184764s (0.28412743798980955s per sentence) memory Δ 7.6 MB 8.8 MB -> 16.4 MB
[2]-th Encoding Time: 2.062344681005925s (0.2577930851257406s per sentence) memory Δ 10.6 KB 16.4 MB -> 16.4 MB
[3]-th Encoding Time: 2.0252770339138806s (0.2531596292392351s per sentence) memory Δ 57.6 KB 16.4 MB -> 16.5 MB
[4]-th Encoding Time: 1.860660734004341s (0.23258259175054263s per sentence) memory Δ 22.9 KB 16.5 MB -> 16.5 MB
[5]-th Encoding Time: 1.8343108689878136s (0.2292888586234767s per sentence) memory Δ 0 Bytes 16.5 MB -> 16.5 MB
[6]-th Encoding Time: 1.8377297670813277s (0.22971622088516597s per sentence) memory Δ 0 Bytes 16.5 MB -> 16.5 MB
[7]-th Encoding Time: 1.7649552969960496s (0.2206194121245062s per sentence) memory Δ 0 Bytes 16.5 MB -> 16.5 MB
[8]-th Encoding Time: 1.7062554890289903s (0.21328193612862378s per sentence) memory Δ 0 Bytes 16.5 MB -> 16.5 MB
[9]-th Encoding Time: 1.7196878109825775s (0.2149609763728222s per sentence) memory Δ 0 Bytes 16.5 MB -> 16.5 MB
==> AVG 0.24220429751148914s per sentence
```

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
numb3r3: @tadejsv I understand what's your point. I think this PR is a temporal one as a hotfix.  Because I need to fix the multi-gpu issue in one of our EAP projects. I think we should take more time to work on this issue. **A better generalized solution should be the future. **

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
tadejsv: I tried running your script, but it crashed - seems my RAM is not enough? I changed the model to `distilbert-base-uncased` and text to `"hello there, how are you doing"`, and did not see any improvement from warmup

```
[0]-th Encoding Time: 0.11161029800132383 s (0.0034878218125413696 s memory Δ 138.4 KB 738.9 KB -> 877.3 KB)
[1]-th Encoding Time: 0.11114512800122611 s (0.003473285250038316 s memory Δ 119.8 KB 877.3 KB -> 997.1 KB)
[2]-th Encoding Time: 0.1059087829926284 s (0.0033096494685196376 s memory Δ 5.5 KB 997.1 KB -> 1002.6 KB)
[3]-th Encoding Time: 0.10730337499990128 s (0.003353230468746915 s memory Δ 0 Bytes 1002.6 KB -> 1002.6 KB)
[4]-th Encoding Time: 0.11572465299104806 s (0.003616395405970252 s memory Δ 0 Bytes 1002.6 KB -> 1002.6 KB)
[5]-th Encoding Time: 0.09653310100839008 s (0.00301665940651219 s memory Δ 0 Bytes 1002.6 KB -> 1002.6 KB)
[6]-th Encoding Time: 0.0972781849995954 s (0.003039943281237356 s memory Δ 264 Bytes 1002.6 KB -> 1002.8 KB)
[7]-th Encoding Time: 0.1003801460028626 s (0.0031368795625894563 s memory Δ 0 Bytes 1002.8 KB -> 1002.8 KB)
[8]-th Encoding Time: 0.09722546499688178 s (0.0030382957811525557 s memory Δ 0 Bytes 1002.8 KB -> 1002.8 KB)
[9]-th Encoding Time: 0.10319713100034278 s (0.003224910343760712 s memory Δ 0 Bytes 1002.8 KB -> 1002.8 KB)
```

tadejsv: In fact, even the improvements you see seem to come only after the 3rd pass. So I would say this is something very task-specific, and should be handled instead by the user (sending a few items through the flow to warm up).

numb3r3: FYI, https://aws.amazon.com/blogs/machine-learning/reduce-ml-inference-costs-on-amazon-sagemaker-for-pytorch-models-using-amazon-elastic-inference/ also include warmup to calcluate the inference latency

numb3r3: Yes, you are right. I think I got an mistake. The warmup should only take effect for `tensorflow`, `torchscript` and `onnx`.  I will trim the warmup code block. Thanks for your comments.

tadejsv: Yes, that's true, you usually do a warmup before performing measurements, just in case.
But if we don't have any strong reason to think that this will always significantly improve performance, I would not include it in the repo. For us this means more maintnance and code that takes more effort to understand, and has more potential to break. And similar as what I said about the multi-gpu thing - if we think this works, we should include it in every pytorch executor, which is a lot of effort. I think the benefit in this case does not outweigh the cost.

As the things are right now, this warmup seems modest, at best, and might require several warmup runs to achieve this efficiency, and whether it helps or not is unclear and depends on the situation.

Perhaps would be good to hear a third opinion here as well.

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
tadejsv: @numb3r3 If it is nececssary, I am not completely against merging it in as a hotfix. But what about the alternative of, for now, copying this code and creating a custom executor in the EAP project? If that is not too feasible we can merge it

jinahub/encoders/text/TransformerTorchEncoder/tests/unit/test_encoder.py
```diff
@@ -11,12 +11,11 @@
 
 
 def test_compute_tokens():
-    enc = TransformerTorchEncoder(base_tokenizer_model="bert-base-cased")
+    enc = TransformerTorchEncoder()
 
     tokens = enc._generate_input_tokens(["hello this is a test", "and another test"])
 
     assert tokens["input_ids"].shape == (2, 7)
-    assert tokens["token_type_ids"].shape == (2, 7)
```
tadejsv, line 19: Good catch!

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
numb3r3: > copying this code and creating a custom executor in the EAP project?

Yes, this can solve my problem. But I still want to encourage our EAP partners to use `jinahub`.

jinahub/encoders/text/TransformerTorchEncoder/tests/data/test_data.txt
```diff
@@ -0,0 +1,50 @@
+The ROMEO website, iOS app and Android app are commonly used by the male gay community to find friends, dates, love or get informed about LGBT+ topics.
```
tadejsv, line 1: Could you remove this file, since it seems it is not used anywhere in the tests? Or am I mistaken?

jinahub/encoders/text/TransformerTorchEncoder/transform_encoder.py
tadejsv: Ok then, let's do this temporary hotfix then. But we should ping @JoanFM once he gets back from vacation to take a look at this, as we would need a permanent solution also

tadejsv (approved):

jinahub/encoders/text/TransformerTorchEncoder/tests/data/test_data.txt
numb3r3: it is needed in  `conftest.py` actually.

PR #20, opened 2021-07-27
feat: sentencizer

cristianmtr:
@cristianmtr 

AFTER MERGING

- add note in old
- archive old repo


jinahub/crafters/Sentencizer/manifest.yml
```diff
@@ -0,0 +1,8 @@
+manifest_version: 1
+name: sentencizer
+alias: Sentencizer
+description: The sentencizer splits text into sentences.
+author: Jina AI Dev-Team (dev-team@jina.ai)
+url: https://github.com/jina-ai/executor-text-sentencizer
```
jakob1111996, line 6: wrong url

scripts/tests.sh
```diff
@@ -21,7 +21,6 @@ if [[ -d "tests/" ]]; then
 
   if [[ -f "tests/requirements.txt" ]]; then
   pip install -r tests/requirements.txt
```
jakob1111996, line 23: don't we need a setup.py here?

jakob1111996:
I am working on this.

scripts/tests.sh
cristianmtr: We shouldn't be using `setup.py` anymore.

See discussions in Slack.

1. Executors are not meant to be pypi packages
2. do relative imports in tests

jakob1111996 (approved):
LGTM

PR #29, opened 2021-07-28
feat: minranker

Yongxuanzhang:


jinahub/rankers/MinRanker/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+
+### 🐳 Via Docker
+
+1. Clone the repo and build the docker image
+
+	```shell
+	git clone https://github.com/jina-ai/executors.git
+	cd jinahub/rankers/
```
cristianmtr, line 67: add subfolder

jinahub/rankers/MinRanker/README.md
```diff
@@ -0,0 +1,98 @@
...
+	```python
+	from jina import Flow
+	
+	f = Flow().add(uses='docker://min-ranker:latest')
+	```
+	
+
+## 🎉️ Example 
+
+Here we **MUST** show a **MINIMAL WORKING EXAMPLE**. We recommend to use `jinahub+docker://MyDummyExecutor` for the purpose of boosting the usage of Jina Hub.
```
cristianmtr, line 82: Add example, just copy-paste simplified test

jinahub/rankers/MinRanker/manifest.yml
```diff
@@ -0,0 +1,8 @@
+manifest_version: 1
+name: MinRanker
```
cristianmtr, line 2: make sure you push to hubble and add the keys to secrets

jinahub/rankers/MinRanker/README.md
Yongxuanzhang: thanks for reminding

jinahub/rankers/MinRanker/manifest.yml
Yongxuanzhang: yes, https://github.com/jina-ai/executors-secrets/pull/3

jinahub/rankers/MinRanker/README.md
Yongxuanzhang: done

jinahub/rankers/MinRanker/README.md
```diff
@@ -0,0 +1,117 @@
...
+	docker build -t min-ranker .
+	```
+
+1. Use `min-ranker` in your codes
+
+	```python
+	from jina import Flow
+	
+	f = Flow().add(uses='docker://min-ranker:latest')
+	```
```
nan-wang, line 77: Check out the changes at #41
```suggestion

```

nan-wang (changes requested):

jinahub/rankers/MinRanker/README.md
jakob1111996: Done

jinahub/rankers/MinRanker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM jinaai/jina:2.0
```
jakob1111996, line 1:
```suggestion
FROM jinaai/jina:master-py38-perf
```

jinahub/rankers/MinRanker/tests/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jina==2.0.10
```
jakob1111996, line 1: I think we should use `git+https://github.com/jina-ai/jina.git@v2.0.10#egg=jina` for the tests

tadejsv: What is the reason for that?

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -0,0 +1,52 @@
...
+    :param default_traversal_paths: traverse path on docs, e.g. ['r'], ['c']
+    :param args:  Additional positional arguments
+    :param kwargs: Additional keyword arguments
+    """
+
+    def __init__(self,
+                 metric: str = None,
+                 default_traversal_paths: List[str] = None,
+                 *args,
+                 **kwargs):
```
tadejsv, line 27: Can you format this file with black? `black -S .`

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -0,0 +1,52 @@
...
+from jina.logging.logger import JinaLogger
+
+
+class MinRanker(Executor):
+    """
+    :class:`MinRanker` aggregates the score
+    of the matched doc from the matched chunks.
+    For each matched doc, the score is aggregated
+    from all the matched chunks belonging to that doc.
+    :param metric: the distance metric used in `scores`
```
tadejsv, line 17: Maybe format this to look nicer - no need to break lines so early, and there should be an empty line between description and args

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self,
+                 metric: str = None,
+                 default_traversal_paths: List[str] = None,
+                 *args,
+                 **kwargs):
+        super().__init__(*args, **kwargs)
+        self.logger = JinaLogger(self.__class__.__name__)
+        if not metric:
+            self.logger.error('metric should not be None')
```
tadejsv, line 31: No need for this at all, see my other comment

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -0,0 +1,52 @@
...
+    For each matched doc, the score is aggregated
+    from all the matched chunks belonging to that doc.
+    :param metric: the distance metric used in `scores`
+    :param default_traversal_paths: traverse path on docs, e.g. ['r'], ['c']
+    :param args:  Additional positional arguments
+    :param kwargs: Additional keyword arguments
+    """
+
+    def __init__(self,
+                 metric: str = None,
```
tadejsv, line 24: If `metric` has to be set, simply do not provide the default and let Python handle missing arguments

tadejsv (changes requested):

jinahub/rankers/MinRanker/minranker.py
Yongxuanzhang: ah yes, thanks!

jinahub/rankers/MinRanker/minranker.py
Yongxuanzhang: I should write the logging msg more clearer,

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -0,0 +1,56 @@
...
+        self,
+        metric: str = None,
+        default_traversal_paths: List[str] = None,
+        *args,
+        **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self.logger = JinaLogger(self.__class__.__name__)
+        if not metric:
+            self.logger.error("metric should not be None, please set it the same as your indexer's metric")
```
tadejsv, line 32: What I mean is that this checking is not necessary at all. If an argument to a function is required, simply do not provide a default value, i.e.

```python
def __init__(self, metric: str):
```
instead of 
```python
def __init__(self, metric: str = None):
```
This way python will throw a very clear error that you need to provide the `metric` argument. There is no need for us to implement it if it is already part of base Python.

Yongxuanzhang: It's not sufficient only to say that we need this param, we should tell users how to set it in logging, I just fixed it. Can you check if that is ok?

Yongxuanzhang: That's my fault not making it clear at first

jinahub/rankers/MinRanker/tests/requirements.txt
jakob1111996: This does not use the pypi package. It instead uses git clone to download the package code.
The advantage is that our CI runs do not pollute the pypi download statistics.
See this Slack post by nan: https://jinaai.slack.com/archives/C0183GCN8AK/p1624870427247100

jakob1111996 (approved):
LGTM

jinahub/rankers/MinRanker/minranker.py
tadejsv: I still don't think this is the way to go. This description of what the metric value should be belongs in the parameter description, not in a warning that appears when the value is not set.

If your concern is that users will set an invalid value, then you would need some kind of validation. But validation  and missing args/defaults are two different things.

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -0,0 +1,56 @@
...
+        self.metric = metric
+        self.default_traversal_paths = default_traversal_paths or ['r']
```
tadejsv, line 34: This is also not needed - simply adjust
```python
default_traversal_paths: List[str] = ['r'],
```
in function definition

tadejsv (changes requested):

jinahub/rankers/MinRanker/minranker.py
tadejsv: Actually, the best version would be
```python
default_traversal_paths: Iterable[str] = ('r'),
```

Yongxuanzhang: Ok, I see. The validation is not so straightforward due to the current design of scores. I will adopt your suggestion, thanks!

tadejsv (approved):
LGTM :)

PR #35, opened 2021-07-29
fix: empty push

cristianmtr:
Fix pushing an empty Executor

### DIFF
--- a/scripts/push.sh
+++ b/scripts/push.sh
@@ -2,6 +2,14 @@
 apt-get update && apt-get install -y jq curl
 
 push_dir=$1
+
+# empty change is detected as home directory
+if [ -z "$push_dir" ]
+then
+      echo "\$push_dir is empty"
+      exit 0
+fi
+
 echo pushing $push_dir
 cd $push_dir
 


tadejsv:
How does an empty executor directory even make it to the `needs.prep-push-bed.outputs.matrix` ?

cristianmtr:
> How does an empty executor directory even make it to the `prep-push-bed.outputs.matrix` ?

It's because of this line https://github.com/jina-ai/executors/blob/11a488acb56865f96a7816d2d9b915a5ff0c2041/scripts/get-all-test-paths.sh#L30

If it's an empty array, it prints an empty string

tadejsv:
Ok, so if I understand correctly, this happens when in fact no executors have been changed (e.g. if only the directory's readme is updated)?

cristianmtr:
> Ok, so if I understand correctly, this happens when in fact no executors have been changed (e.g. if only the directory's readme is updated)?

Yes, or other non-executor files

tadejsv (approved):
LGTM!

PR #41, opened 2021-07-29
docs: move other usage to details

cristianmtr:
- remove setup.py
- remove mention of using the Executor on its own
- add note to prereqs that they are only needed if you download the class and use it on its own


nan-wang (dismissed):
LGTM👍

tadejsv (approved):
LGTM!

jakob1111996:
Why do some READMEs have a line `DETAILSSTART` at the end and some others don't?

tadejsv (changes requested):
Good point @jakob1111996 , it seems that's an error - also examples in those cases are deleted. This will need to be fixed

jakob1111996:
> 
> 
> Good point @jakob1111996 , it seems that's an error - also examples in those cases are deleted. This will need to be fixed

How is it supposed to be? Remove the `DETAILSSTART` and add the examples section again?

tadejsv:
> > Good point @jakob1111996 , it seems that's an error - also examples in those cases are deleted. This will need to be fixed
> 
> How is it supposed to be? Remove the `DETAILSSTART` and add the examples section again?

Yes, it should only be deleted up to the examples section.

jakob1111996:
> 
> 
> > > Good point @jakob1111996 , it seems that's an error - also examples in those cases are deleted. This will need to be fixed
> > 
> > 
> > How is it supposed to be? Remove the `DETAILSSTART` and add the examples section again?
> 
> Yes, it should only be deleted up to the examples section.

Alright, I can handle that

jakob1111996:
By the way, where are the other installation options located now?

tadejsv:
Thanks @jakob1111996 ! It seems we are removing all other install options to steer people to use jinahub. On the readme of this repo you still have the pip install (not very well done, as it does not mention requirements), and docker install options.

I am fine with this, as it helps keep the base template simple, and not overload people with options.

tadejsv (commented):
Great job fixing stuff here, @jakob1111996 

Since you're already working on it, can I ask you to also delete these html comments that are not used anymore:

```
<!-- START doctoc generated TOC please keep comment here to allow auto update -->
<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
```
and
```
<!-- END doctoc generated TOC please keep comment here to allow auto update -->
```

tadejsv (approved):
LGTM

PR #43, opened 2021-07-30
feat: adapt _rank / _aggregate logic to SimpleIndexer

winstonww:
The problem we currently have in one of our eaps is that given a set of index docs with encoded chunks and a set of query docs with encoded chunks, we want to match the top level query docs with the top level index docs (For a simple use case that illustrates the problem, please see the new test added). Adapting the `_rank` function from one of our eaps, the PR adds a simple `_aggregate` method to our `SimpleIndexer` to achieve this goal. Note, this is currently only a draft to demo the problem. The method currently only accommodates one traversal pattern (i.e. `['c']`). Need to figure out a way to make this work with various traversal patterns. Publishing the PR to gather more thoughts. Thank you.

### DIFF
--- a/jinahub/indexers/SimpleIndexer/__init__.py
+++ b/jinahub/indexers/SimpleIndexer/__init__.py
@@ -1,8 +1,11 @@
+from collections import defaultdict
 from typing import Dict, Tuple, Optional, List
 
+from pathlib import Path
 import numpy as np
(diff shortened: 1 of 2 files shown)


winstonww:
Closing this PR as we want to separate ranking logic from `SimpleIndexer`.

PR #45, opened 2021-07-31
feat: add huggingface clipimage transformers

makram93:
Hi @tadejsv,

I am raising the PR. Waiting for your feedback


makram93:
I ran this test locally and it passed: CLIPImageEncoder/tests/unit/test_clip.py::test_openai_embed_match 
However, it is failing in CI check. Why is that happening?

jinahub/encoders/image/CLIPImageEncoder/clip_image.py
```diff
@@ -1,82 +1,95 @@
...
+                                "[H, W, C], in the RGB format (C=3),"
+                                f" but got C={doc.blob.shape[2]} instead."
+                            )
+                        images.append(doc.blob)
+
+                    tensor = self.pre_processor(images=images, return_tensors="pt")
+                    tensor = tensor.to(self.device)
+                    embeddings = self.model.get_image_features(**tensor)
+                    embeddings = embeddings.cpu().numpy()
+                    for idx, doc in enumerate(batch_docs):
```
JoanFM, line 94: u can use zip

JoanFM (dismissed):
Please keep `use_default_preprocessing` option for now

makram93:
@JoanFM, 
I have added the `use_default_preprocessing` param in my latest commit.

@tadejsv,
I have noticed that in hugginface implementation they perform `transforms.ToTensor()` after `center_crop` transformation. However, in your `AudioCLIPImageEncoder`, you have set 'ToTensor()' as first step in `_default_transform`. This does not impact the semantic test but the embedding values obtained are different. This happens because `ToTensor` scales the value between [0, 1] using the entire pixels whereas when used after `center_crop`, it only considers the cropped pixels for scaling. A small example below, in which if we perform `ToTensor` as the first operation then the embeddings don't match the original embeddings. 
``` python
    _IMAGE_SIZE = 224
    _IMAGE_MEAN = 0.48145466, 0.4578275, 0.40821073
    _IMAGE_STD = 0.26862954, 0.26130258, 0.27577711

    from PIL import Image
    import requests
    from transformers import CLIPProcessor, CLIPModel

    url = "http://images.cocodataset.org/val2017/000000039769.jpg"
    image = Image.open(requests.get(url, stream=True).raw)

    model = CLIPModel.from_pretrained("openai/clip-vit-base-patch32")

    # Using processor
    processor = CLIPProcessor.from_pretrained("openai/clip-vit-base-patch32")
    inputs = processor(images=image, return_tensors="pt")
    outputs_p = model.get_image_features(**inputs).detach().cpu().numpy()

    # Using transforms with ToTensor after resize
    default_transforms = transforms.Compose(
        [
            transforms.ToPILImage(),
            transforms.Resize(_IMAGE_SIZE, interpolation=Image.BICUBIC),
            transforms.CenterCrop(_IMAGE_SIZE),
            transforms.ToTensor(),
            transforms.Normalize(_IMAGE_MEAN, _IMAGE_STD),
        ]
    )
    images = default_transforms(np.array(image)).unsqueeze(dim=0)
    outputs_t1 = model.get_image_features(images).detach().cpu().numpy()

    # Using transforms with ToTensor in the first step
    default_transforms = transforms.Compose(
        [
            transforms.ToTensor(),
            transforms.Resize(_IMAGE_SIZE, interpolation=Image.BICUBIC),
            transforms.CenterCrop(_IMAGE_SIZE),
            transforms.Normalize(_IMAGE_MEAN, _IMAGE_STD),
        ]
    )
    images = default_transforms(image).unsqueeze(dim=0)
    outputs_t2 = model.get_image_features(images).detach().cpu().numpy()

    np.testing.assert_almost_equal(outputs_p, outputs_t2, 5)
```

jinahub/encoders/image/CLIPImageEncoder/README.md
```diff
@@ -1,107 +1,123 @@
...
+The following parameters can be passed on initialization:
+- `pretrained_model_name_or_path`: Can be either:
+    - A string, the model id of a pretrained CLIP model hosted
+        inside a model repo on huggingface.co, e.g., 'openai/clip-vit-base-patch32'
+    - A path to a directory containing model weights saved, e.g., ./my_model_directory/
+- `use_default_preprocessing`: Whether to use the default preprocessing on
+        images (blobs) before encoding them. If you disable this, you must ensure
+        that the images you pass in have the correct format, see the ``encode`` method
+        for details.
+- `device`: device that the model is on (should be "cpu", "cuda" or "cuda:X",
```
tadejsv, line 16: Also add back default preprocessing arg heere

jinahub/encoders/image/CLIPImageEncoder/README.md
```diff
@@ -1,107 +1,123 @@
...
+#### using docker images
+
+Use the prebuilt images from JinaHub in your Python code: 
+
+```python
+from jina import Flow
+	
+f = Flow().add(
+        uses='jinahub+docker://CLIPImageEncoder',
+        volumes='/your_home_folder/.cache/clip:/root/.cache/clip')
```
tadejsv, line 52: This volume is not needed anymore

jinahub/encoders/image/CLIPImageEncoder/README.md
```diff
@@ -1,107 +1,123 @@
...
+```
+
+or in the `.yml` config.
+	
+```yaml
+jtype: Flow
+pods:
+  - name: encoder
+    uses: 'jinahub+docker://CLIPImageEncoder'
+    volumes: '/your_home_folder/.cache/clip:/root/.cache/clip'
```
tadejsv, line 62: Also not needed

jinahub/encoders/image/CLIPImageEncoder/README.md
```diff
@@ -1,107 +1,123 @@
...
+```
+
+#### using source code
+Use the source code from JinaHub in your Python code:
+
+```python
+from jina import Flow
+	
+f = Flow().add(uses='jinahub://CLIPImageEncoder',
```
tadejsv, line 71:
```suggestion
f = Flow().add(uses='jinahub://CLIPImageEncoder')

```

jinahub/encoders/image/CLIPImageEncoder/README.md
```diff
@@ -1,107 +1,123 @@
...
+**CLIPImageEncoder** is a class that wraps the image embedding functionality using the **CLIP** model from huggingface transformers.
+
+The **CLIP** model originally was proposed in [Learning Transferable Visual Models From Natural Language Supervision](https://cdn.openai.com/papers/Learning_Transferable_Visual_Models_From_Natural_Language_Supervision.pdf).
+
+The following parameters can be passed on initialization:
+- `pretrained_model_name_or_path`: Can be either:
+    - A string, the model id of a pretrained CLIP model hosted
+        inside a model repo on huggingface.co, e.g., 'openai/clip-vit-base-patch32'
+    - A path to a directory containing model weights saved, e.g., ./my_model_directory/
+- `use_default_preprocessing`: Whether to use the default preprocessing on
```
tadejsv, line 12: Also, can you add a parameter here `base_feature_extractor` - exactly the same logic as `base_tokenizer` in the text version

tadejsv (changes requested):
As always, great job :)

I've left some comments - the main thing is to use CLIPFeatureExtractor for preprocessing.

Regarding your commend on order of transforms - where do you think this comes up?
The failure of the tests before was only for one image, so I don't think it was this. But in any case I will check this again after you apply the changes.

tadejsv:
One more comment here: from what I see HF will only divide by 255 and transform to float if the image is a PIL image, if it is a numpy image it will assume this is already the case. So you should also do that manually in case of default preprocessing.

And the `ToTensor` transform actually also does that (divide by 255), so this is the source of the differences in your example.

makram93:
Hi @tadejsv ,

I checked the `__call__` method in `CLIPFeatureExtractor` . It performs [resize](https://huggingface.co/transformers/_modules/transformers/models/clip/feature_extraction_clip.html#CLIPFeatureExtractor.resize) and [center crop](https://huggingface.co/transformers/_modules/transformers/models/clip/feature_extraction_clip.html#CLIPFeatureExtractor.center_crop) on PIL Image only. If it is `nd.array` then it converts it to PIL Image. After that it performs [normalization](https://huggingface.co/transformers/_modules/transformers/image_utils.html#ImageFeatureExtractionMixin.normalize) during which it converts the PIL Image to a numpy array and divides it by 255. So I think we don't need to manually divide numpy image before feeding it to  `CLIPFeatureExtractor`

tadejsv:
@makram93 We need to do the division by 255, because our documents contain `blob`s - which are numpy arrays and not PIL images. Same reason for needing to switch the channel axis to the first place

makram93:
@tadejsv I understand that we have numpy arrays in `blobs` but if we pass it through `CLIPFeatureExtractor` preprocessor then it will automatically handle it by first converting it to PIL and then later dividing it by 255 before normalising it.

I am little confused now. This is what I have understood so far. Correct me if I'm wrong. We keep two args `base_feature_extractor` and `use_default_preprocessing` in `clip_image.py`. 
- If `use_default_preprocessing=True`, then we accept arrays with uint or float as dtype. In this sceanrio we use `CLIPFeatureExtractor` to preprocess it. We don't need to shift axis as [`to_pil_image`](https://huggingface.co/transformers/_modules/transformers/image_utils.html#ImageFeatureExtractionMixin.to_pil_image) handles it. Neither we need to divide it by 255 as [`to_numpy_array`](https://huggingface.co/transformers/_modules/transformers/image_utils.html#ImageFeatureExtractionMixin.to_numpy_array) handles it
- If `use_default_preprocessing=False`, then we assume preprocessing has already been done and we simply create a tensor out of it and pass it to the model to get embeddings.

tadejsv:
> @tadejsv I understand that we have numpy arrays in `blobs` but if we pass it through `CLIPFeatureExtractor` preprocessor then it will automatically handle it by first converting it to PIL and then later dividing it by 255 before normalising it.

I see you are right, this happens in resize and center crop. Then good, we don't need to do it actually.

> I am little confused now. This is what I have understood so far. Correct me if I'm wrong. We keep two args `base_feature_extractor` and `use_default_preprocessing` in `clip_image.py`.
> 
> * If `use_default_preprocessing=True`, then we accept arrays with uint or float as dtype. In this sceanrio we use `CLIPFeatureExtractor` to preprocess it. We don't need to shift axis as [`to_pil_image`](https://huggingface.co/transformers/_modules/transformers/image_utils.html#ImageFeatureExtractionMixin.to_pil_image) handles it. Neither we need to divide it by 255 as [`to_numpy_array`](https://huggingface.co/transformers/_modules/transformers/image_utils.html#ImageFeatureExtractionMixin.to_numpy_array) handles it
> * If `use_default_preprocessing=False`, then we assume preprocessing has already been done and we simply create a tensor out of it and pass it to the model to get embeddings.

Yes, all correct

makram93:
Hi @tadejsv,

I have pushed the code with updated changes.

nan-wang:
@makram93 I'm afraid the test is failing https://github.com/jina-ai/executors/pull/45/checks?check_run_id=3232579039#step:4:415

jinahub/encoders/image/CLIPImageEncoder/tests/unit/test_clip.py
```diff
@@ -1,86 +1,262 @@
...
+    docs.match(docs)
+    matches = ["cat", "dog", "helicopter", "airplane"]
+    for i, doc in enumerate(docs):
+        assert doc.matches[1].id == matches[i]
+
+
+def test_openai_embed_match():
+    data_dir = Path(__file__).parent.parent / "imgs"
+    dog = Document(id="dog", blob=np.array(Image.open(data_dir / "dog.jpg")))
+    cat = Document(id="cat", blob=np.array(Image.open(data_dir / "cat.jpg")))
```
tadejsv, line 237: Just remove the cat. I did some exploring, and the difference is in how HF and CLIP (torchvision transforms) implement the resize + centercrop transformations. The difference is subtle - i saved both images with only these 2 transformations and they looked visually the same, yet their pixel values were different.

Not sure why exactly this is, but in any case there is nothing that can be done about this, nor do I think it should be a major problem.

tadejsv (changes requested):

makram93:
@tadejsv done

tadejsv (approved):
@makram93 Good job!!

jinahub/encoders/image/CLIPImageEncoder/clip_image.py
```diff
@@ -1,82 +1,128 @@
...
+    :param default_batch_size: fallback batch size in case there is no batch size sent in the request
+    :param default_traversal_paths: fallback traversal path in case there is no traversal path sent in the request
+    """
+
+    def __init__(
+        self,
+        pretrained_model_name_or_path: str = "openai/clip-vit-base-patch32",
+        base_feature_extractor: str = "openai/clip-vit-base-patch32",
+        use_default_preprocessing: bool = True,
+        device: Optional[str] = "cpu",
```
tadejsv, line 35:
```suggestion
        device: str = "cpu",
```

tadejsv (approved):
Some tidbits

tadejsv (approved):
Ready to merge

PR #46, opened 2021-08-02
fix: numpy searcher docs traversal

Yongxuanzhang:
the NumpySearcher used to add matches to the root level, but it should add them into the corresponding traversal level, then will be helpful for ranker to rerank


tadejsv:
I'm fine with this - can you add some tests?

Yongxuanzhang:
> I'm fine with this - can you add some tests?

ah yes. We need tests

cristianmtr (approved):

Yongxuanzhang:
> > I'm fine with this - can you add some tests?
> 
> ah yes. We need tests, will add soon

tadejsv (approved):
LGTM

PR #47, opened 2021-08-02
refactor: use docs match function and update SimpleIndexer to use latest Jina version

winstonww:


jinahub/indexers/SimpleIndexer/__init__.py
```diff
@@ -80,7 +81,7 @@ def search(self, docs: 'DocumentArray', parameters: Dict, **kwargs):
         for _q, _ids, _dists in zip(flat_docs, idx, dist):
             for _id, _dist in zip(_ids, _dists):
                 d = Document(self._docs[int(_id)], copy=True)
-                d.scores['cosine'] = 1 - _dist
+                d.scores[self.distance_metric] = 1 - _dist
```
JoanFM, line 84: This does not make much sense for `euclidean` distance then?

Also let's use `match` interface from `DocumentArray`

jinahub/indexers/SimpleIndexer/__init__.py
```diff
@@ -57,8 +58,8 @@ def index(self, docs: 'DocumentArray', parameters: Dict, **kwargs):
         :param docs: the docs to add
         :param parameters: the parameters dictionary
         """
-        traversal_path = parameters.get('traversal_paths', self.default_traversal_paths)
-        flat_docs = docs.traverse_flat(traversal_path)
+        traversal_paths = parameters.get('traversal_paths', self.default_traversal_paths)
```
JoanFM, line 61: Let's make this code robust to having `None` as `docs` inputs since `docs` is an Optional argument

JoanFM (changes requested):

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: Addressed, please see [here](https://github.com/jina-ai/executors/pull/47/commits/cfcfe9cfdea788b19822ba1ab53f480f7667c143) for the change.

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: Ah true, thanks for pointing that out. I have updated the search function to use `docs.match`. Since we want the score/similarity in this case, what do you think of using `lambda: q_emb, d_emb: -self.distance(q_emb, d_emb)` as the distance metric?

nan-wang (changes requested):
Please update the following codes as well.
https://github.com/jina-ai/executors/blob/3c5d786f3565a4864ac9bf66b9cd567c6701040c/jinahub/indexers/SimpleIndexer/manifest.yml#L6
https://github.com/jina-ai/executors/blob/3c5d786f3565a4864ac9bf66b9cd567c6701040c/jinahub/indexers/SimpleIndexer/Dockerfile#L1

cristianmtr:
Is this https://github.com/jina-ai/jina/issues/3109 ? If so ,can u link it and assign the ticket to yourself?

jinahub/indexers/SimpleIndexer/__init__.py
```diff
@@ -44,60 +44,28 @@ def __init__(
...
 
     @requests(on='/search')
-    def search(self, docs: 'DocumentArray', parameters: Dict, **kwargs):
+    def search(self, docs: Optional['DocumentArray'] = None, parameters: Optional[Dict] = {}, **kwargs):
         """Perform a vector similarity search and retrieve the full Document match
 
         :param docs: the Documents to search with
         :param parameters: the parameters for the search"""
-        traversal_path = parameters.get('traversal_paths', self.default_traversal_paths)
+        if not docs: return
```
numb3r3, line 65: Why the `traversal_path` block is removed in this commit?

numb3r3 (changes requested):

winstonww:
@cristianmtr Yes, I believe they are the same. This PR refactors `SimpleIndexer` so that it uses the new `match` function of `DocumentArray`. Correct me if I am wrong, though there are other new methods implemented for `DocumentArray`, I believe the `match` function is the only we need for now in `SimpleIndexer`.

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: Ah yes, `traversal_path` should not be removed, I have now reverted this change [here](https://github.com/jina-ai/executors/pull/47/commits/413b5527cae3e780d90f76203a3c0fdd2faf0462).

winstonww:
@nan-wang I have updated the documentation link in `manifest.yml`. As the change in this PR is compatible with the latest Jina version (`2.0.17`), I will update the `Dockerfile` version as soon as `2.0.17` is released.

jinahub/indexers/SimpleIndexer/__init__.py
```diff
@@ -44,60 +44,30 @@ def __init__(
...
         """All Documents to the DocumentArray
         :param docs: the docs to add
         :param parameters: the parameters dictionary
         """
-        traversal_path = parameters.get('traversal_paths', self.default_traversal_paths)
-        flat_docs = docs.traverse_flat(traversal_path)
+        if not docs: return
+        traversal_paths = parameters.get('traversal_paths', self.default_traversal_paths)
+        flat_docs = docs.traverse_flat(traversal_paths)
         self._docs.extend(flat_docs)
```
numb3r3, line 56: Does it work with `flat_docs` is empty?

jinahub/indexers/SimpleIndexer/__init__.py
```diff
@@ -44,60 +44,30 @@ def __init__(
...
-    def search(self, docs: 'DocumentArray', parameters: Dict, **kwargs):
+    def search(self, docs: Optional['DocumentArray'] = None, parameters: Optional[Dict] = {}, **kwargs):
         """Perform a vector similarity search and retrieve the full Document match
 
         :param docs: the Documents to search with
         :param parameters: the parameters for the search"""
-        traversal_path = parameters.get('traversal_paths', self.default_traversal_paths)
+        if not docs: return
+        traversal_paths = parameters.get('traversal_paths', self.default_traversal_paths)
+        flat_docs = docs.traverse_flat(traversal_paths)
```
numb3r3, line 67: Same problem with `flat_docs = []`

numb3r3 (changes requested):

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: Addressed, please see [here](https://github.com/jina-ai/executors/pull/47/commits/f9fe67dfa7b6b8958a85e1dd5c5298b697fca959) for the change. Also added tests for this case, thank you.

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: Addressed, please see here for the change. Also added tests for this case.

nan-wang (dismissed):
LGTM👍

winstonww:
There are a couple of outstanding issues still need to be addressed, will push another commit soon.

jinahub/indexers/SimpleIndexer/tests/test_simple_indexer.py
```diff
@@ -48,3 +80,118 @@ def test_simple_indexer(tmpdir):
     assert search_docs_id[0].embedding is None
     indexer.fill_embedding(search_docs_id)
     assert search_docs_id[0].embedding is not None
+
+
+def test_simple_indexer_loading(tmpdir):
+    pass
```
winstonww, line 86: Will complete this in next commit.

winstonww: Completed.

numb3r3 (dismissed):
LGTM

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: The negative sign is now removed from the metric function. I have also doubled check places where `SimpleIndexer` is used to ensure the usage is correct after the modification.

jinahub/indexers/SimpleIndexer/__init__.py
```diff
@@ -44,60 +44,34 @@ def __init__(
...
 
     @requests(on='/search')
-    def search(self, docs: 'DocumentArray', parameters: Dict, **kwargs):
+    def search(self, docs: Optional['DocumentArray'] = None, parameters: Optional[Dict] = {}, **kwargs):
         """Perform a vector similarity search and retrieve the full Document match
 
         :param docs: the Documents to search with
         :param parameters: the parameters for the search"""
-        traversal_path = parameters.get('traversal_paths', self.default_traversal_paths)
+        if not docs or not self._docs: return
```
JoanFM, line 65: if not self._docs at share time maybe it would be good to have some logs?

JoanFM (changes requested):

jinahub/indexers/SimpleIndexer/__init__.py
winstonww: Addressed, please see [here](https://github.com/jina-ai/executors/pull/47/commits/a154dffeefe6112776c32d6d0a206d163af1bf66) for the change.

winstonww:
Rebased onto latest main branch. Unit test failures are probably due to merge conflicts, currently investigating.

winstonww:
Merge conflicts are now resolved, also updated Dockerfile.

nan-wang (approved):
LGTM👍

PR #50, opened 2021-08-02
feat: add DRPReaderRanker

tadejsv:


jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,203 @@
...
+                        titles,
+                    ) = self._get_outputs(doc.text, contexts, titles)
+
+                answer_spans += spans
+                answer_relevance_scores += relevance_scores
+                answer_span_scores += span_scores
+                answer_titles += titles
+
+            # Make sure answers are sorted by relevance scores
+            sorting_list = list(zip(answer_relevance_scores, answer_span_scores))
```
JoanFM, line 165: Are we sure this is needed? It seemed before it was already granted. So it would mean we can skip this step

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,203 @@
...
+            answer_spans = _sort_by(answer_spans, sorting_list)
+
+            answer_titles = _sort_by(answer_titles, sorting_list)
+            answer_relevance_scores = _sort_by(answer_relevance_scores, sorting_list)
+            answer_span_scores = _sort_by(answer_span_scores, sorting_list)
```
JoanFM, line 170: Even if it is needed, we are doing 4 sorts. We could sort only once and avoid zipping and only keep the valid indices

JoanFM (changes requested):

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def _batcher(iterable, n=1):
+    """Batch an iterable. Here temporarily until
+    https://github.com/jina-ai/jina/issues/3068
+    is solved.
+    """
+    l = len(iterable)
+    for ndx in range(0, l, n):
+        yield iterable[ndx : min(ndx + n, l)]
```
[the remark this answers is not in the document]
tadejsv: Yes you're right

JoanFM (changes requested):
title seems to be an Optional input, so I would not add this extra requirement. Make it optional

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,203 @@
...
+    :param default_traversal_paths: Default traversal paths for processing documents,
+        used if the traversal path is not passed as a parameter with the request.
+    :param device: The device (cpu or gpu) that the model should be on.
+    """
+
+    def __init__(
+        self,
+        pretrained_model_name_or_path: str = 'facebook/dpr-reader-single-nq-base',
+        base_tokenizer_model: Optional[str] = None,
+        title_tag_key: str = 'title',
```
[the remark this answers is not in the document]
tadejsv: Ok, will try to do that

jinahub/rankers/DPRReaderRanker/dpr_reader.py
tadejsv: It is needed bc of batching. Within the batch matches are sorted, but across batches.

JoanFM: Ok this makes sense.

Then store these in some structure that u can sort only once. Or sort by keeping the valid indices to access them later. But having 4 sorts is an overkill

JoanFM (commented):
sorting a zip may also work by the first element

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,203 @@
...
+                        tags={'title': title, 'span_score': span_score},
+                    )
+                )
+            doc.matches.extend(new_matches)
+
+    def _get_outputs(
+        self, question: str, contexts: List[str], titles: List[str]
+    ) -> Tuple[List[float], List[float], List[str], List[str]]:
+
+        encoded_inputs = self.tokenizer(
```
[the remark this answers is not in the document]
tadejsv: I could try batching accross questions (root docs) - I am just worried that this would lead to code becoming too unreadable

jinahub/rankers/DPRReaderRanker/dpr_reader.py
tadejsv: Ok I'll replace it

jinahub/rankers/DPRReaderRanker/dpr_reader.py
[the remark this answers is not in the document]
JoanFM: It would definitely be less nice, but it will be definitely more powerful

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+            ``batch_size``. For example::
+            parameters={'traversal_paths': ['r'], 'batch_size': 10}
+        """
+        if not docs:
+            return None
+
+        new_matches = []
+        old_matches = []
+
+        docs_list = []
```
JoanFM, line 120: docs_list I do not think is needed. If u simply use traverse_flat and store it u already have it

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+
+            for rel_score, span, tags, ref_ind in zip(*outputs):
+                scores = {'relevance_score': rel_score}
+                new_match = Document(text=span, scores=scores, tags=tags)
+                new_matches.append(new_match)
+                new_match_ref_doc_dict[new_match.id] = ref_ind
+
+        # Clear old matches
+        for doc in docs_list:
+            doc.matches.clear()
```
JoanFM, line 156: U can clear the old matches in the first for loop

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+        for ind, doc in enumerate(docs.traverse_flat(trav_paths)):
+            if not doc.text:
+                raise ValueError(
+                    f'No question (text) found for document with id {doc.id}.'
+                    ' DPRReaderRanker requires a question from the main document'
+                    ' and context (text) from its matches.'
+                )
+
+            docs_list.append(doc)
+            old_matches += doc.matches
```
JoanFM, line 136: It is very weird to flatten matches inside a list with matches from different questions

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+            for match in doc.matches:
+                old_match_ref_doc_dict[match.id] = ind
+
+        # Traverse all matches
+        batch_size = parameters.get('batch_size', self.default_batch_size)
```
JoanFM, line 141: I think batching should apply to questions not to matches

JoanFM (changes requested):

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
+from itertools import groupby
+from typing import Dict, Iterable, List, Optional, Tuple, Union
+
+import numpy as np
+import torch
+from jina import Document, DocumentArray, Executor, requests
+from transformers import DPRReader, DPRReaderTokenizerFast
+
+
+def _batcher(iterable, n=1):
```
JoanFM, line 10: batcher functionality is already standarized from jina commons, u can use that one

JoanFM (changes requested):
Sorry for my wrong feedback in a previous review, but looking at the tokenizer documentation from huggingface, it does not seem right to handle many questions at the same time. So previous approach was better and clearer

jinahub/rankers/DPRReaderRanker/tests/unit/test_ranker.py
```diff
@@ -0,0 +1,176 @@
...
+
+    docs = DocumentArray()
+    for question in questions[: request.param]:
+        doc = Document(text=question)
+        matches = [
+            Document(text=passage, tags={'title': title})
+            for passage, title in zip(passages[:10], titles)
+        ]
+        doc.matches.extend(matches)
+        docs.append(doc)
```
nan-wang, line 44: Add some randomness to show the reranking actually works.
```suggestion
        doc.matches.shuffle(seed=0)
        docs.append(doc)
```

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+        # Perpare a list of docs to add matches to later and extract their
+        # matches, plus build a reference to their reference docs
+        trav_paths = parameters.get('traversal_paths', self.default_traversal_paths)
+        for ind, doc in enumerate(docs.traverse_flat(trav_paths)):
+            if not doc.text:
+                raise ValueError(
+                    f'No question (text) found for document with id {doc.id}.'
+                    ' DPRReaderRanker requires a question from the main document'
+                    ' and context (text) from its matches.'
+                )
```
nan-wang, line 133: I suggest skipping the docs without `text`. Instead of raising an exception and stop working, we can raise a warning and continue processing the others.
```suggestion
            if not doc.text:
                raise ValueError(
                    f'No question (text) found for document with id {doc.id}.'
                    ' DPRReaderRanker requires a question from the main document'
                    ' and context (text) from its matches.'
                )
```

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+            num_spans=self.num_spans_per_match * len(questions),
+            num_spans_per_passage=self.num_spans_per_match,
+        )
+        relevance_scores = _logistic_fn([span.relevance_score for span in best_spans])
+        spans = [span.text for span in best_spans]
+        answer_ref_inds = [ref_inds[span.doc_id] for span in best_spans]
+
+        tags = []
+        for span in best_spans:
+            tags_span = {'span_score': _logistic_fn(span.span_score)}
```
nan-wang, line 226: Instead of storing `span_score`, I think putting it into `scores['span_score']` is more intuitive.

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+    is solved.
+    """
+    l = len(iterable)
+    for ndx in range(0, l, n):
+        yield iterable[ndx : ndx + n]
+
+
+def _logistic_fn(x: np.ndarray) -> List[float]:
+    """Compute the logistic function"""
+    return (np.exp(x) / (1 + np.exp(x))).tolist()
```
nan-wang, line 22: The old implementation has extra computation and does not work for 
```python
_logistic_fn(np.array([np.inf, 1, 2]))
```

```suggestion
def _logistic_fn(x: np.ndarray) -> List[float]:
    """Compute the logistic function"""
    return (1 / (1 + np.exp(-x))).tolist()
```

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+            return None
+
+        new_matches = []
+        old_matches = []
+
+        docs_list = []
+        old_match_ref_doc_dict = {}
+        new_match_ref_doc_dict = {}
+
+        # Perpare a list of docs to add matches to later and extract their
```
nan-wang, line 124:
```suggestion
        # Prepare a list of docs to add matches and extract their
```

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+            questions.append(docs_list[old_match_ref_doc_dict[match.id]].text)
+
+        titles = None
+        if self.title_tag_key:
+            try:
+                titles = [match.tags[self.title_tag_key] for match in matches]
+            except KeyError:
+                raise ValueError(
+                    f'All matches are required to have the {self.title_tag_key} tag,'
+                    ' but found somem atches without it.'
```
nan-wang, line 191:
```suggestion
                    ' but found some matches without it.'
```

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # For each context, extract num_spans_per_match best spans
+        best_spans = self.tokenizer.decode_best_spans(
+            encoded_inputs,
+            outputs,
+            num_spans=self.num_spans_per_match * len(questions),
+            num_spans_per_passage=self.num_spans_per_match,
+        )
+        relevance_scores = _logistic_fn([span.relevance_score for span in best_spans])
+        spans = [span.text for span in best_spans]
```
nan-wang, line 221: `span.start_index` and `span.end_index` can be saved into the `matches[i].location`. This would be useful for the visualization purpose.

nan-wang (changes requested):
Good job!

jinahub/rankers/DPRReaderRanker/dpr_reader.py
tadejsv: No, this won't work because they are indices of start and end tokens, not characters.

jinahub/rankers/DPRReaderRanker/dpr_reader.py
tadejsv: Alright

jinahub/rankers/DPRReaderRanker/tests/unit/test_ranker.py
tadejsv: There is already a test that checks that proper matches get to the first place (and matches are constructed from scratch anyways), so it's not needed

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -0,0 +1,192 @@
...
+                needs_attr='text',
+            )
+            for matches in match_batches_generator:
+                inputs = self._prepare_inputs(doc, matches)
+                with torch.no_grad():
+                    new_matches += self._get_new_matches(*inputs)
+
+            # Make sure answers are sorted by relevance scores
+            new_matches.sort(
+                key=lambda x: (x.scores['relevance_score'], x.scores['span_score']),
```
JoanFM, line 137: I think it should be x.scores['relevance_score'].value

tadejsv: You are right, these are some typos I didn't notice (will commit a fix in the next few mins)

nan-wang (approved):
LGTM👍

PR #53, opened 2021-08-03
fix: packaging

cristianmtr:
- fix setup.py
- add init.py where missing (required in order for modules to be installed)

Classes will be available as:

```
from jinahub.indexers.SimpleIndexer import SimpleIndexer
from jinahub.encoders.image.ImageTorchEncoder.torch_encoder import ImageTorchEncoder
```


nan-wang (approved):
LGTM👍

PR #55, opened 2021-08-04
fix: rename init files, remove hubble files

tadejsv:
- renames the `__init__.py` files (that were used as the main file of the module)
- does some minor cleanup (like removing the now-redundant `hubble.yml` files).
- standardizes the `__init__.py` to be empty files (some few ones had some top-level imports, I removed that as this is *not* meant to be a package).


cristianmtr:
`__init__` is required in order to perform imports. Is this change really necessary?

tadejsv:
@cristianmtr the point is to rename the files that contain executor code to `__init__.py`, not to get rid of them (we need either that or `setup.py` for testing anyway)

tadejsv:
This is ready for review - failures in CI are failures of executors themselves, not related and out of scope for this PR

cristianmtr:
> This is ready for review - failures in CI are failures of executors themselves, not related and out of scope for this PR

Why? The tests were passing before, right?

cristianmtr (changes requested):
please try to fix the tests too. We can't merge otherwise

tadejsv:
The failure occurs because of lines such as this https://github.com/jina-ai/executors/blob/14fcaab0663fbc54c00a0c1805167d5051d1d0a7/jinahub/indexers/storage/PostgreSQLStorage/tests/test_postgres_dbms.py#L54

I see this is due to https://github.com/jina-ai/jina/pull/3058, which removed context managament from Document

cristianmtr (approved):
thanks for fixing

PR #58, opened 2021-08-05
feat: remove numpy

cristianmtr:
Closes https://github.com/jina-ai/jina/issues/3110

AFTER APPROVAL / BEFORE MERGING

on Hubble

- remove Numpy
- remove Compounds with Numpy
- push new Compounds
- add secrets


jinahub/indexers/README.md
```diff
@@ -21,7 +21,7 @@ See below figure for how this would look like:
 
 ![](./.github/img/replicas.png)
 
-In the above case, the Storage could be the [PostgreSQL](storage/PostgreSQLStorage)-based Storage, while the Query Flow could be based on [NumpyPostgresSearcher](searcher/compound/NumpyPostgresSearcher).
+In the above case, the Storage could be the [PostgreSQL](storage/PostgreSQLStorage)-based Storage, while the Query Flow could be based on [NumpyPostgresSearcher](searcher/compound/FaissPostgresSearcher).
```
tadejsv, line 24:
```suggestion
In the above case, the Storage could be the [PostgreSQL](storage/PostgreSQLStorage)-based Storage, while the Query Flow could be based on [FaissPostgresSearcher](searcher/compound/FaissPostgresSearcher).
```

tadejsv:
Are you still having problems here with non-consistent results?

cristianmtr:
> Are you still having problems here with non-consistent results?

Yes, that's why the check is only `>`

Yongxuanzhang (commented):
If we remove the searcher, do we also want to remove it from bubble? @cristianmtr

cristianmtr:
> If we remove the searcher, do we also want to remove it from bubble? @cristianmtr

Yes, good point. We should. Once this gets approved, I'll remove it, along with the compounds, and add these two new Compounds

nan-wang (approved):
LGTM👍

PR #64, opened 2021-08-09
fix(posgres): fix py-modules and import error

numb3r3:


jinahub/indexers/storage/PostgreSQLStorage/postgres.py
```diff
@@ -8,7 +8,7 @@
 
 from jina_commons import get_logger
 from jina_commons.indexers.dump import export_dump_streaming
-from .postgreshandler import PostgreSQLHandler
+from postgreshandler import PostgreSQLHandler
```
tadejsv, line 11: Why is this needed?

tadejsv:
I also tried fixing this - seems that nothing apart from renaming the `postgres.py` file back to `__init__.py` works. 

@numb3r3 , any idea why this is? Also, why does Jina need to modify python's import system (which works without problems for all other purposes) at all? What prevents us from just using normal python imports? Sorry if these questions are too broad, it's just something I have a hard time understanding

numb3r3:
> why does Jina need to modify python's import system (which works without problems for all other purposes) at all? 

To support to load executor from a `YAML` config file (i.e., `exc = Executor.load_config('config.yml')`), we **dynamicly import** the modules one by one included in `py-modules`.  

> What prevents us from just using normal python imports?

I think the problem we want to address in this PR is not caused by our jina `import` machinery. The relative import failure also happens on other normal python projects. For example,

```
- my_module
    - __init__.py  # empy content
    - a.py. # import `from .foo import *`
    - foo.py
```
```
# the following line will raise `relative import` error
python a.py 
```

tadejsv:
@numb3r3 no this is caused precisely by the jina import machinery. Everything works when you are testing locally - this means that jina imports modules in a different way then python.

tadejsv:
Actually, you are right. Then the question is why does this work in local tests

Issue #61, opened 2021-08-08
Postgres - wrong path 


florian-hoenicke:
Steps to reproduce:
```
docker build -t postgres .
docker run -it postgres 
```

Logs:
```
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMWWWMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMWNNNNNNNWMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMNNNNNNNNNMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMWNNNNNNNNMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMWNNNWWMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMWxxxxxxxxxOMMMMMNxxxxxxxxx0MMMMMKddddddxkKWMMMMMMMMMMMMXOxdddxONMMMM
MMMMMMMMMMMMXllllllllldMMMMM0lllllllllxMMMMMOllllllllllo0MMMMMMMM0olllllllllo0MM
MMMMMMMMMMMMXllllllllldMMMMM0lllllllllxMMMMMOlllllllllllloWMMMMMdllllllllllllldM
MMMMMMMMMMMMXllllllllldMMMMM0lllllllllxMMMMMOllllllllllllloMMMM0lllllllllllllllK
MMMMMMMMMMMMKllllllllldMMMMM0lllllllllxMMMMMOllllllllllllllKMMM0lllllllllllllllO
MMMMMMMMMMMMKllllllllldMMMMM0lllllllllxMMMMMOllllllllllllll0MMMMollllllllllllllO
[...]

PR #69, opened 2021-08-09
fix: psql import

cristianmtr:
- revert PSQL init


tadejsv (changes requested):
If you are attaching change to CI to this PR, then we should do it properly and implement the change that does not require this reversion of the previous PR, see here https://github.com/jina-ai/executors/issues/61#issuecomment-895203429

cristianmtr:
I'll revert the CI from this PR

tadejsv:
I think to get the tests to work you need to put things back into `__init__.py`. With changing the CI I was just refering to addition of Docker testing (which I think is a great idea, just better suited for a separate PR, if we want this as a quickfix). 

For the error we were getting we don't actually need Docker to reproduce, this will be enough (I think?)
```python
def test_config():
    ex = BaseExecutor.load_config("../../config.yml")
```

tadejsv (approved):
Ok, if you tested locally (with docker) and it works, let's merge

PR #77, opened 2021-08-11
fix: use top k

slettner:


tadejsv (dismissed):
Let's add a test for it

nan-wang (approved):
LGTM👍

PR #78, opened 2021-08-11
fix: match bug

jakob1111996:
Fix some minor bugs in the SimpleIndexer and MinRanker which crashed the examples.

### DIFF
--- a/jinahub/indexers/SimpleIndexer/simple_indexer.py
+++ b/jinahub/indexers/SimpleIndexer/simple_indexer.py
@@ -36,6 +36,7 @@ class SimpleIndexer(Executor):
         self._docs = DocumentArrayMemmap(self.workspace + f'/{index_file_name}')
         self.default_traversal_paths = default_traversal_paths or ['r']
         self.default_top_k = default_top_k
+        self.metric_name = distance_metric
         if distance_metric == 'cosine':
             self.distance = _cosine
         elif distance_metric == 'euclidean':
@@ -97,6 +98,7 @@ class SimpleIndexer(Executor):
                 _ext_A(_norm(q_emb)), _ext_B(_norm(d_emb))
             ),
             limit=top_k,
+            metric_name=self.metric_name
         )
         self._flush = False
 
--- a/jinahub/indexers/SimpleIndexer/simple_indexer.py
+++ b/jinahub/indexers/SimpleIndexer/simple_indexer.py
@@ -94,9 +94,7 @@ class SimpleIndexer(Executor):
         top_k = int(parameters.get('top_k', self.default_top_k))
         flat_docs.match(
             self._docs,
-            metric=lambda q_emb, d_emb, _: self.distance(
-                _ext_A(_norm(q_emb)), _ext_B(_norm(d_emb))
-            ),
+            metric=lambda q_emb, d_emb, _: self.distance(q_emb, d_emb),
             limit=top_k,
             metric_name=self.metric_name
         )
@@ -131,8 +129,8 @@ def _ext_B(B):
     return B_ext
 
 
-def _euclidean(A_ext, B_ext):
-    sqdist = A_ext.dot(B_ext).clip(min=0)
+def _euclidean(A, B):
+    sqdist = _ext_A(A).dot(_ext_B(B)).clip(min=0)
     return np.sqrt(sqdist)
 
 
@@ -141,4 +139,4 @@ def _norm(A):
 
 
 def _cosine(A_norm_ext, B_norm_ext):
-    return A_norm_ext.dot(B_norm_ext).clip(min=0) / 2
+    return _ext_A(_norm(A)).dot(_ext_B(_norm(B))).clip(min=0) / 2
--- a/jinahub/indexers/SimpleIndexer/simple_indexer.py
+++ b/jinahub/indexers/SimpleIndexer/simple_indexer.py
@@ -138,5 +138,5 @@ def _norm(A):
     return A / np.linalg.norm(A, ord=2, axis=1, keepdims=True)
 
 
-def _cosine(A_norm_ext, B_norm_ext):
+def _cosine(A, B):
     return _ext_A(_norm(A)).dot(_ext_B(_norm(B))).clip(min=0) / 2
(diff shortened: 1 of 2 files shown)


jakob1111996:
I will fix the tests until the end of the week.

jinahub/rankers/MinRanker/minranker.py
```diff
@@ -22,11 +22,16 @@ def __init__(
...
+        metric_is_distance: bool = False,
         *args,
         **kwargs
     ):
         super().__init__(*args, **kwargs)
         self.metric = metric
+        if metric_is_distance:
+            self.distance_mult = 1
+        else:
+            self.distance_mult = -1
```
nan-wang, line 34: What is the motivation to add this part? I think the user should be aware of this when they choose the index. Therefore this should be done in the indexer part.

nan-wang (changes requested):

jinahub/indexers/SimpleIndexer/simple_indexer.py
```diff
@@ -93,10 +94,11 @@ def search(
         top_k = int(parameters.get('top_k', self.default_top_k))
         flat_docs.match(
             self._docs,
-            metric=lambda q_emb, d_emb, _: self.distance(
+            metric=lambda q_emb, d_emb: self.distance(
                 _ext_A(_norm(q_emb)), _ext_B(_norm(d_emb))
```
numb3r3, line 98: It looks like an error. The `_norm` should not be applied for the `euclidean` metric.

jinahub/rankers/MinRanker/minranker.py
numb3r3: > Therefore this should be done in the indexer part.

@nan-wang Does that mean the matched score resulted from indexer should always be a distance (or similarity) value? I don't think we can guarantee this assumption.

numb3r3 (changes requested):

jinahub/indexers/SimpleIndexer/simple_indexer.py
```diff
@@ -93,10 +94,11 @@ def search(
         top_k = int(parameters.get('top_k', self.default_top_k))
         flat_docs.match(
             self._docs,
-            metric=lambda q_emb, d_emb, _: self.distance(
+            metric=lambda q_emb, d_emb: self.distance(
```
winstonww, line 97: I believe the metric function in the latest jina version (2.0.17) actually requires 3 arguments, which explains why the unit test is currently failing.

jinahub/indexers/SimpleIndexer/simple_indexer.py
```diff
@@ -93,10 +94,11 @@ def search(
...
                 _ext_A(_norm(q_emb)), _ext_B(_norm(d_emb))
             ),
             limit=top_k,
+            metric_name=self.metric_name
```
winstonww, line 101: Is it possible that you can point me to where this `metric_name` is needed in examples?

winstonww (changes requested):
Thanks for the PR, just some questions from my side.

jinahub/indexers/SimpleIndexer/simple_indexer.py
JoanFM: match is to be handled like in #79

JoanFM (changes requested):

jinahub/indexers/SimpleIndexer/simple_indexer.py
jakob1111996: Yes I had some kind of a version conflict so - this is no longer required - i will remove it

jinahub/indexers/SimpleIndexer/simple_indexer.py
jakob1111996: The problem we have here is that we pass a lambda function to the .match() method as a distance metric. Therefore, the name of the metric is not 'cosine' as expected but <lambda> instead. Thie means I have to retrieve the score like so: `doc.scores['<lambda>]'` which is not at all what we want.

jinahub/rankers/MinRanker/minranker.py
jakob1111996: I was a bit worried here because we actually assume that all metrics are distance metrics. We can not use scores (the higher the better) at all currently.

jakob1111996: If this is a design choice, that is also fine for me. Then I only expect distance metrics here :smile:

jakob1111996: If things are a distance metric, we need to switch the sign in the sorting operations, however.

jinahub/indexers/SimpleIndexer/simple_indexer.py
jakob1111996: I agree.

winstonww (approved):
LGTM

JoanFM:
I actually do not think this PR makes any more sense

jakob1111996:
> I actually do not think this PR makes any more sense

That is not right. I misinterpreted some details but this PR still fixes two bugs:
1. The name of the scores when using the lambda function is incorrect. It needs to be retrieved by `doc.scores[<lambda>]` instead of scores[cosine]
2. The min ranker sorts the wrong way around (it currently expects scores not distances)

JoanFM:
> > I actually do not think this PR makes any more sense
> 
> That is not right. I misinterpreted some details but this PR still fixes two bugs:
> 
> 1. The name of the scores when using the lambda function is incorrect. It needs to be retrieved by `doc.scores[<lambda>]` instead of scores[cosine]
> 2. The min ranker sorts the wrong way around (it currently expects scores not distances)

The lambda function is not used anymore

jakob1111996:
> > > I actually do not think this PR makes any more sense
> > 
> > 
> > That is not right. I misinterpreted some details but this PR still fixes two bugs:
> > 
> > 1. The name of the scores when using the lambda function is incorrect. It needs to be retrieved by `doc.scores[<lambda>]` instead of scores[cosine]
> > 2. The min ranker sorts the wrong way around (it currently expects scores not distances)
> 
> The lambda function is not used anymore

Then somebody else must have fixed that already. Then, only the minranker bug remains. I will create a new PR for that.

PR #79, opened 2021-08-12
feat: use properly match with no custom distance func

JoanFM:


jinahub/indexers/SimpleIndexer/simple_indexer.py
```diff
@@ -36,22 +35,21 @@ def __init__(
...
         self.default_top_k = default_top_k
-        if distance_metric == 'cosine':
-            self.distance = _cosine
-        elif distance_metric == 'euclidean':
-            self.distance = _euclidean
-        else:
-            raise ValueError('This distance metric is not available!')
+        self._distance = distance_metric
+        self._use_scipy = True
+        if distance_metric in ['cosine', 'euclidean', 'sqeuclidean']:
```
cristianmtr, line 40: Can u throw an error if distance is not supported?

JoanFM: we do not have an exhaustive list. It depends on what `scipy` supports.

JoanFM: These ones are the ones supported natively, there are others offered via `scipy`

JoanFM: I changed the docstrings

winstonww (approved):
LGTM

PR #82, opened 2021-08-13
fix: min ranker becomes simple ranker

jakob1111996:
The min ranker was not really a min ranker previously.
We did sort the elements of the array descendingly, this means the maximum score is in the first place.
This PR fixes this bug and repairs the Minranker to rank by minimum score (minimum distance).


Yongxuanzhang (changes requested):
This is not a bug, more like a naming issue, I think we can provide parameters to users to choose use min, max, or mean, same as the previous simple aggregate ranker. Thanks!

Yongxuanzhang:
Thanks for pointing it out. 😄

jakob1111996:
> Thanks for pointing it out. smile

The name of this executor is minranker - if we also allow mean and max ranking, we should probably rename it right?

nan-wang (dismissed):
LGTM👍

nan-wang:
>  if we also allow mean and max ranking, we should probably rename it right?

Yes, maybe `SimpleRanker`?

jinahub/rankers/SimpleRanker/simpleranker.py
```diff
@@ -21,12 +23,15 @@ class MinRanker(Executor):
     def __init__(
         self,
         metric: str = 'cosine',
+        ranking: str = 'min',
```
Yongxuanzhang, line 26: this name is not so clear to me. But I'm not sure what could be a better naming. maybe ranking_function, aggregrate_function..

jinahub/rankers/SimpleRanker/tests/unit/test_ranker.py
```diff
@@ -0,0 +1,98 @@
...
+            assert match.tags
+            assert (
+                match.scores['cosine'].value
+                <= doc.matches[i + 1].scores['cosine'].value
+            )
+
+
+@pytest.mark.parametrize('default_traversal_paths', [['r'], ['c']])
+def test_max_ranking(documents_chunk, documents_chunk_chunk, default_traversal_paths):
+    ranker = SimpleRanker(metric='cosine', ranking='max',
```
Yongxuanzhang, line 32: maybe parametrize the ranking ['min','max'...]?

jinahub/rankers/SimpleRanker/simpleranker.py
jakob1111996: I personally prefer ranking over the longer renking_function. I will write a few more comments so it is more clear.

Yongxuanzhang (approved):
👍

Yongxuanzhang (approved):

PR #72, opened 2021-08-09
fix: testing and relative imports fix

tadejsv:
Stuff to do

- [x] Import executor in `__init__.py`
- [x] Use only relative imports
- [x] Fix VGGish
- [x] Fix AudioCLIPEncoder
- [x] Add small load from config test to each executor
- [x] add CI to test docker image start @cristianmtr 
- [x] test manually that using relative imports and importing only `__init__.py` works for `jinahub://`

PR #83, opened 2021-08-13
fix: use iterator to load dumped data

numb3r3:
- use an `iterator` to fetch data from a large-scale dumped file
- remove `self._vecs` to reduce the memory usages, so as to support large index
- remove `fill_embedding` request handler as `self._vecs` is removed

### DIFF
--- a/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
+++ b/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
@@ -65,7 +65,7 @@ class FaissSearcher(Executor):
         dump_path: Optional[str] = None,
         prefech_size: Optional[int] = None,
         default_traversal_paths: List[str] = ['r'],
-        is_distance: bool = True,
+        is_distance: bool = False,
         default_top_k: int = 5,
         on_gpu: bool = False,
         *args,
@@ -261,7 +261,10 @@ class FaissSearcher(Executor):
 
             normalize_L2(vecs)
         dists, ids = self.index.search(vecs, top_k)
-        
+
+        if self.metric == 'inner_product':
+            dists = 1 - dists
+
         for doc_idx, matches in enumerate(zip(ids, dists)):
             for m_info in zip(*matches):
                 idx, dist = m_info
--- a/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
+++ b/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
@@ -261,8 +261,7 @@ class FaissSearcher(Executor):
 
             normalize_L2(vecs)
         dists, ids = self.index.search(vecs, top_k)
-        if self.metric == 'inner_product':
-            dists = 1 - dists
+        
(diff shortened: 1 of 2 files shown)


jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -27,6 +27,7 @@ class FaissSearcher(Executor):
         The data will only be loaded if `requires_training` is set to True.
     :param max_num_training_points: Optional argument to consider only a subset of training points to training data from `train_filepath`.
         The points will be selected randomly from the available points
+    :param prefech_size: the number of data to pre-load into RAM memory
```
cristianmtr, line 30:
```suggestion
    :param prefech_size: the number of data to pre-load into RAM
```

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -88,13 +91,25 @@ def __init__(
...
+            )
+            self._ids = np.array(list(ids_iter))
             self._ext2int = {v: i for i, v in enumerate(self._ids)}
-            self._vecs = np.array(list(vecs))
-            self.num_dim = self._vecs.shape[1]
-            self.dtype = self._vecs.dtype
-            self.index = self._build_index(self._vecs)
+
+            self._prefech_data = []
+            if self.prefech_size and self.prefech_size > 0:
```
cristianmtr, line 101: So if you have a dump of size N and prefetch of size X (X < N), how do you build the entire N index? It seems here that you're limiting to just X

cristianmtr: ah no, nvm, got it

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -62,6 +63,7 @@ def __init__(
         normalize: bool = False,
         nprobe: int = 1,
         dump_path: Optional[str] = None,
+        prefech_size: Optional[int] = None,
```
cristianmtr, line 66: This should be a non-zero int by default, right? `None` doesn't help

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: This is only the first prefetch step. The full fetch procedure happens in `self._build_partial_index` function

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: here, `None` means load the full data at once. Yes, `None` will disable prefetch.

JoanFM (approved):

cristianmtr:
Don't merge until https://github.com/jina-ai/executors/pull/72

nan-wang (approved):
LGTM👍

PR #95, opened 2021-08-17
fix: add traversal paths

Yongxuanzhang:
as a user the images may be in chunks level, so we better support this

### DIFF
--- a/jinahub/crafters/ImageNormalizer/normalizer.py
+++ b/jinahub/crafters/ImageNormalizer/normalizer.py
@@ -2,7 +2,7 @@ __copyright__ = "Copyright (c) 2021 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
 from pydoc import locate
-from typing import Tuple, Union, Iterable, List
+from typing import Tuple, Union, Iterable
 import numpy as np
 import PIL.Image as Image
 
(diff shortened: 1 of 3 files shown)


jinahub/crafters/ImageNormalizer/normalizer.py
```diff
@@ -45,22 +47,24 @@ def __init__(
         else:
             self.target_dtype = target_dtype
 
-
     @requests
-    def craft(self, docs: DocumentArray, **kwargs) -> DocumentArray:
-        filtered_docs = DocumentArray(
-            list(filter(lambda d: 'image/' in d.mime_type, docs))
-        )
+    def craft(self, docs: DocumentArray, parameters, **kwargs) -> DocumentArray:
```
nan-wang, line 51:
```suggestion
    def craft(self, docs: DocumentArray, parameters: dict, **kwargs) -> DocumentArray:
```

jinahub/crafters/ImageNormalizer/normalizer.py
```diff
@@ -45,22 +47,24 @@ def __init__(
...
+        if docs:
```
nan-wang, line 52:
```suggestion
        if docs is None:
            return
        ...
```

nan-wang (dismissed):

numb3r3 (approved):
LGTM

cristianmtr (approved):

PR #96, opened 2021-08-17
feat: psql import

cristianmtr:
- adding an import from PSQL (to avoid dumping to disk stage)
- PSQL distributes data to virtual shards, to prepare for shard allocation
- snapshotting the table state
- getting the document changes since timestamp
- soft deletion in PSQL
- [x] add cleanup endpoint. And make cleanup happen before snapshot
- [x] implement full syncing via delta
- [x] more parametrization at init


jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -90,40 +104,62 @@ def __init__(
...
                 dump_path, str(self.runtime_args.pea_id)
             )
-            self._ids = np.array(list(ids_iter))
-            self._doc_id_to_offset = {v: i for i, v in enumerate(self._ids)}
+            iterator = zip(ids_iter, vecs_iter)
+        elif dump_func is not None:
+            iterator = dump_func(shard_id=self.runtime_args.pea_id)
+        else:
+            self.logger.warning(
+                'No "dump_data" or "dump_func" passed to "FaissIndexer".'
```
maximilianwerk, line 133:
```suggestion
                'No "dump_path" or "dump_func" passed to "FaissIndexer".'
```

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -91,49 +111,62 @@ def __init__(
         self._prefetch_data = []
 
         self.logger = get_logger(self)
+        self._load_dump(dump_path, dump_func, prefetch_size, **kwargs)
+
+    def _load_dump(self, dump_path, dump_func, prefetch_size, **kwargs):
```
maximilianwerk, line 116: This function became to complex and should be split into sub-functions. Same is true for `_train`. Actually, I think we should redesign this a bit. Let me have some thinking about it.

jinahub/indexers/searcher/FaissSearcher/tests/requirements.txt
```diff
@@ -1,4 +1 @@
-jina==2.0.21
-faiss-cpu==1.6.5
```
maximilianwerk, line 2: Just out of interest: Why don't we need this anymore?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,51 +2,122 @@
...
 try:
-    from jinahub.indexers.storage.PostgreSQLStorage import (
-        PostgreSQLStorage,
-    )
-except:
-    from jina_executors.indexers.storage.PostgreSQLStorage import (
-        PostgreSQLStorage,
-    )
+    from jinahub.indexers.storage.PostgreSQLStorage import PostgreSQLStorage
+except:  # noqa: E722
```
maximilianwerk, line 20: Why not catch `ImportError`?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,51 +2,122 @@
...
+                f'for {name}. Use .rolling_update() to re-initialize...'
+            )
+
+        if use_dump_func:
             self._kv_indexer = PostgreSQLStorage(**kwargs)
+            dump_func = self._kv_indexer.get_snapshot
+            self._vec_indexer = FaissSearcher(dump_func=dump_func, **kwargs)
         else:
+            self._kv_indexer = PostgreSQLStorage(**kwargs)
+            self._vec_indexer = FaissSearcher(dump_path=dump_path, **kwargs)
```
maximilianwerk, line 73: `self._vec_indexer, self._kv_indexer = self._init_indexers(...)` will reduce complexity here.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,51 +2,122 @@
...
+
+    @requests(on='/sync')
+    def sync(self, parameters: Dict, **kwargs):
+        """
+        Sync the data from the PSQLStorage into the FaissSearcher
+        """
+        use_delta = parameters.get('use_delta', False)
+        only_delta = parameters.get('only_delta', False)
```
maximilianwerk, line 81: Why having `only_delta` here?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,51 +2,122 @@
...
+    :param use_dump_func: whether to use the dump
+     function of PostgreSQLStorage, when dump_path is not provided
+    """
 
     def __init__(
         self,
-        dump_path=None,
+        dump_path: Optional[str] = None,
+        use_dump_func: bool = False,
+        total_shards: int = None,
```
maximilianwerk, line 38: Either this becomes `Optional[int]` or needs another default. I prefer having default `= 1`.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,51 +2,122 @@
...
-                f'No dump path provided for {self}. Use .rolling_update() to re-initialize...'
+                'Syncing via delta method. This cannot guarantee consistency'
             )
+            pass
+            # we get the timestamp from time 0 and pass an argument to Indexer so
+            # that it can ignore the "delete"s
+
+        else:
+            self.logger.info('Syncing via snapshot...')
+            if self.total_shards:
```
maximilianwerk, line 93: This `if` becomes obsolete, if we default `total_shards` to `1`, or?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,51 +2,122 @@
...
+                timestamp = self._kv_indexer.last_snapshot_timestamp
+                self._vec_indexer = FaissSearcher(
+                    dump_func=dump_func, **self._init_kwargs
+                )
+
+                if use_delta:
+                    self.logger.info(f'Now adding delta from timestamp {timestamp}')
+
+                    deltas = self._kv_indexer._get_delta(
+                        shard_id=0, total_shards=3, timestamp=timestamp
```
maximilianwerk, line 106: Old code? I guess this needs to be some variable and not hard-coded `0` and `3`.

jinahub/indexers/storage/PostgreSQLStorage/postgres_indexer.py
```diff
@@ -57,6 +60,7 @@ def __init__(
         self.database = database
         self.table = table
         self.logger = get_logger(self)
+        self.total_shards = total_shards
```
maximilianwerk, line 63: This is a semantically different total_shards, then in the `FaissPostgresSearcher`. Can we have a better naming?

jinahub/indexers/storage/PostgreSQLStorage/postgres_indexer.py
```diff
@@ -177,3 +199,53 @@ def search(self, docs: DocumentArray, parameters: Dict, **kwargs):
...
+        if shard_id == total_shards - 1:
+            shards_to_get = vshards[
+                shard_id
+                * vshard_part : ((shard_id + 1) * vshard_part + vshard_remainder)
+            ]
+        else:
+            shards_to_get = vshards[
+                shard_id * vshard_part : (shard_id + 1) * vshard_part
+            ]
+        return [str(shard_id) for shard_id in shards_to_get]
```
maximilianwerk, line 234: This should be a `@staticmethod`, getting `self.total_shards` (however this will be renamed) as an input as well. Then it is super easy to heavily test it, since nobody is able to get, whether this function is right without tests :p

maximilianwerk (changes requested):
Please read first all comments before addressing them. Some of them are connected.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -91,49 +111,62 @@ def __init__(
...
                 for _ in range(prefetch_size):
                     try:
-                        self._prefetch_data.append(next(vecs_iter))
-
+                        self._prefetch_data.append(next(iterator))
                     except StopIteration:
                         break
             else:
-                self._prefetch_data = list(zip(ids_iter, vecs_iter))
+                self._prefetch_data = list(iterator)
```
maximilianwerk, line 148: Why is this prefetching happening here anyhow? The only reason is to get `self.num_dim` and `self.dtype` filled. This seems a quite weak reason to do so. This can happen in the very first `next` call inside `_iterate_vectors_and_save_ids` now. This is much cleaner.

jinahub/indexers/searcher/FaissSearcher/tests/requirements.txt
cristianmtr: They are installed by the main `requirements.txt`.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: Yep, of course.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: We need to get the `get_snapshot` of `kv_indexer` to pass to the faiss. So this wouldn't work, right?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: This is the method for not using snapshotting but just getting all the data from timestamp 0. 

Is is an alternative to the normal snapshot + delta. Where should it go?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: This needs to be passed from core, I suggest. I will make it optional and leave the warning there to stand out until solved.

florian-hoenicke: can we assume the total number of shards stays the same? I think in the short-term yes, but long-term we might want to scale the number of shards up and down

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: total_shards needs to be known all the time

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: Yes, fixed now

jinahub/indexers/storage/PostgreSQLStorage/postgres_indexer.py
cristianmtr: Yes, renamed to virtual shards

jinahub/indexers/storage/PostgreSQLStorage/postgres_indexer.py
cristianmtr: Yes, good point. Done

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
cristianmtr: Yes, idk. @numb3r3 ?

cristianmtr: Can be addressed in a future PR maybe.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: Stays the same for all systems, or what do you mean? By default it's `None` unless you pass it.

numb3r3 (changes requested):

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
```diff
@@ -201,28 +235,63 @@ def update(self, docs: DocumentArray, *args, **kwargs):
...
         :param kwargs: other keyword arguments
         :return record: List of Document's id after deletion
         """
         cursor = self.connection.cursor()
-        psycopg2.extras.execute_batch(
-            cursor,
-            f'DELETE FROM {self.table} WHERE doc_id = %s;',
-            [(doc.id,) for doc in docs],
-        )
+        if soft_delete:
```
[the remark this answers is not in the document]
cristianmtr: We can tell if a record is deleted because the 'emb' and 'doc' columns are NULL. I think that's sufficient, right?

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
```diff
@@ -260,23 +329,110 @@ def get_size(self):
         records = cursor.fetchall()
         return records[0][0]
 
+    def _get_next_shard(self):
```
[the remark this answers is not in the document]
cristianmtr: What's the advantage? I've tested this approach pretty thoroughly and it works.

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
numb3r3: I don't think so. Assigning `emb` and `doc` as 'NULL' implies erasing the data from the database. That's not soft delete as described in wiki https://en.wiktionary.org/wiki/soft_deletion

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
cristianmtr: Ok, maybe we're using the term incorrectly, but that is what we needed.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
maximilianwerk: I just meant to put all the code from line 60-73 into another function, which return the `kv_indexer` and `vec_indexer`

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
maximilianwerk: It took me another round of reading the code to really understands, what `only_delta` means. But now I got it. Thanks.

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
maximilianwerk: The advantage is, that different indexers would come to the same conclusion to which shard to write to. It is deterministic with only the Document id and number of total shards as input.

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
cristianmtr: Ok, maybe we're using the term incorrectly, but that is what we needed.

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
cristianmtr: So the advantage in our case would be the removal of the extra column, `shard`?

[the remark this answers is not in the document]
cristianmtr: Is this hashing function used by mongo available as a package, or do we need to re-implement?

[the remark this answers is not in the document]
maximilianwerk: Not sure, whether I would remove this column. I think the column should still be there, for a faster dump-generation. Just the calculation should do some hashing instead of round-robin.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
maximilianwerk: This total number of shards should be determined at the startup and never be `None`. And if it is None, it should default to `1` or so. This will make all downstream code easier (e.g. the `             if self.total_shards: ` below.

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
numb3r3: we can simple employ `hashlib` to implement 

```
import hashlib
def hash_fn(key, total_shards):
    sha = hashlib.sha256()
    sha.update(bytes(key, 'utf-8'))
    return int(sha.hexdigest(), 16) % total_shards
```
this is quite simple, but it works

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
numb3r3: We actually need **real soft delete** to make it possible to recover the deleted data from the database.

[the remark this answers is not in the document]
cristianmtr: But we don't need that. The soft deletion is for correctly updating a running FaissSearcher, namely for deleting the entries that have been deleted from PSQL

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
cristianmtr: But won't this affect speed?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: I refactored this a bit. Check again

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -48,81 +52,110 @@ def __init__(
...
+                self._vec_indexer._add_delta(deltas)
+        else:
+            self.logger.warning(
+                'total_shards is None, rolling update '
+                'via PSQL import will not be possible.'
+            )
+
+    def _sync_only_delta(self, parameters, **kwargs):
+        timestamp = parameters.get('timestamp', None)
+        startup = kwargs.get('startup', False)
```
maximilianwerk, line 128: I think this should be part of the parameters. If someone sends a request with `parameters={'startup': True}` it's their fault.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -48,81 +52,110 @@ def __init__(
...
+        if timestamp is None and startup:
+            timestamp = datetime.datetime.min
+
+        if timestamp is None:
+            self.logger.error(
+                f'No timestamp provided in parameters: '
+                f'{parameters}. Cannot do sync delta'
+            )
+            return
```
maximilianwerk, line 137:
```suggestion
        if timestamp is None:
            if startup:
                timestamp = datetime.datetime.min
            else:
                self.logger.error(
                    f'No timestamp provided in parameters: '
                    f'{parameters}. Cannot do sync delta'
                )
                return
```

jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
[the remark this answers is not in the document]
maximilianwerk: I am not sure, about which point in time we are talking about here. I mean, that you do this hashing during a `/index` call. So only once, when storing the Document the very first time to PSQL. So a drop-in replacement of your the current call to `_get_next_shard`.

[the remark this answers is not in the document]
cristianmtr: done, adapted

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,50 +2,172 @@
...
+from jina_commons import get_logger
 
 try:
     from jinahub.indexers.searcher.FaissSearcher import FaissSearcher
-except:
-    from jina_executors.indexers.searcher.FaissSearcher.faiss_searcher import FaissSearcher
+except ImportError:  # noqa: E722
+    from jina_executors.indexers.searcher.FaissSearcher.faiss_searcher import (
+        FaissSearcher,
+    )
```
tadejsv, line 17: This whole thing, plus I guess the installations around it, can now be removed because jina does not use `jinahub` anymore, so there will be no more conflicts.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,50 +2,172 @@
...
+            self.sync(parameters=startup_sync_args)
+
+    def _init_executors(self, dump_path, kwargs, startup_sync, startup_sync_args):
+        kv_indexer = PostgreSQLStorage(**kwargs)
+
+        if startup_sync_args is None and startup_sync:
+            startup_sync_args = {}
+        # if the user passes args for syncing, they must've meant to sync as well
+        if startup_sync_args and (startup_sync is False or startup_sync is None):
+            startup_sync = True
```
maximilianwerk, line 79: This code smells. It cries, that we should only have `startup_sync_args` and not `startup_sync`. And `startup_sync_args` is also ugly, since it is not clear what to provide. Let's have a session tomorrow to see, how we can make this more user-friendly. I have a design in mind. We will do this in a follow-up PR.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -2,50 +2,172 @@
...
+        dump_path: Optional[str] = None,
+        startup_sync: bool = False,
+        startup_sync_args: Optional[None] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
-        # when constructed from rolling update the dump_path is passed via a runtime_arg
+        self.logger = get_logger(self)
+
+        self.total_shards = self.runtime_args.parallel
```
florian-hoenicke, line 47: When using Kubernetes, parallel and replicas will always be set to 1 for each deployed execuror. For now I would suggest to create a new parameter `total_shards` instead.

florian-hoenicke (changes requested):

nan-wang (dismissed):

jinahub/indexers/storage/PostgreSQLStorage/postgres_indexer.py
```diff
@@ -112,10 +126,24 @@ def update(self, docs: DocumentArray, parameters: Dict, **kwargs):
         with self.handler as postgres_handler:
             postgres_handler.update(docs.traverse_flat(traversal_paths))
 
+    @requests(on='/delete')
```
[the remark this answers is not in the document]
cristianmtr: No, cleanup is another endpoint.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: image

jinahub/indexers/searcher/compound/FaissPostgresSearcher/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM jinaai/jina:master-py37-perf as base
 
 # install and upgrade pip
 RUN apt-get update && apt-get install -y git
+RUN python3.7 -m pip install --upgrade pip
```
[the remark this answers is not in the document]
cristianmtr: Removed.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: Ah yes, good catch. I added some code to handle that.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
maximilianwerk: Much nicer :+1:

jinahub/indexers/searcher/FaissSearcher/tests/integration/test_train_index.py
```diff
@@ -79,3 +80,19 @@ def test_train_and_index(metas, tmpdir):
...
+@pytest.mark.gpu
+@pytest.mark.docker
+def test_docker_runtime_gpu(build_docker_image_gpu: str):
+    with pytest.raises(subprocess.TimeoutExpired):
+        subprocess.run(
+            [
+                'jina',
+                'executor',
+                f'--uses=docker://{build_docker_image_gpu}',
+                '--gpus all',
```
tadejsv, line 94: This should be `'--gpus', 'all'` - would you mind also fixing the documentation for it? We fixed this yesterday but forgot to update documentation

cristianmtr: Done

maximilianwerk (approved):

PR #100, opened 2021-08-17
docs: add warning to simple indexer

jakob1111996:
The simple indexer is currently experiencing a bug when running it with docker `jinahub+docker://SimpleIndexer`.
After closing the docker container, it sometimes happens that the stored memmap which is mounted into the container is corrupted. After data, the entire index is broken and the user needs to index from scratch.

The error does not happen when using `jinahub://SimpleIndexer` which is why I quickly changed the README to make the user aware of this problem.

### DIFF
--- a/jinahub/indexers/SimpleIndexer/README.md
+++ b/jinahub/indexers/SimpleIndexer/README.md
@@ -23,9 +23,11 @@
 
 ### 🚚 Via JinaHub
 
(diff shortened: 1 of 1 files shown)


tadejsv:
Any idea on how/why this corruption happens?

nan-wang (approved):
LGTM👍

PR #99, opened 2021-08-17
fix: optimize psql dump test

cristianmtr:
Closes https://github.com/jina-ai/internal-tasks/issues/152
Part of https://github.com/jina-ai/internal-tasks/issues/142

BEFORE MERGING: 

merge commons PR https://github.com/jina-ai/jina-commons/pull/26
and remove https://github.com/jina-ai/executors/pull/99/files#diff-27c5febcc9ce93e8e5341f263b7d68962733b8a2a86c583723d025a1db39d7a3L1

PR #101, opened 2021-08-17
feat: some performance improvements

maximilianwerk:
some performance improvements for the dump and the generation of documents for the test.


maximilianwerk:
Already addressed in #99

PR #103, opened 2021-08-17
feat: add delete and update for simple indexer

Yongxuanzhang:


jinahub/indexers/SimpleIndexer/simple_indexer.py
```diff
@@ -97,6 +97,45 @@ def search(
...
+        if docs is None:
+            return
+        traversal_paths = parameters.get(
+            'traversal_paths', self.default_traversal_paths
+        )
+        flat_docs = docs.traverse_flat(traversal_paths)
+        delete_docs_ids = set(flat_docs.get_attributes('id'))
+        for idx in reversed(range(len(self._docs))):
+            if self._docs[idx].id in delete_docs_ids:
+                del self._docs[idx]
```
numb3r3, line 116: I think the  `__delete__` method in `DAM` can use `doc.id` directly
```suggestion
        for doc in flat_docs:
            if doc.id is None: continue
            deleted_doc = self._docs.get(doc.id, None)
            if deleted_doc:
                del self._docs[doc.id]
        
```

jinahub/indexers/SimpleIndexer/simple_indexer.py
```diff
@@ -97,6 +97,45 @@ def search(
...
+            'traversal_paths', self.default_traversal_paths
+        )
+        flat_docs = docs.traverse_flat(traversal_paths)
+
+        update_docs_ids = set(flat_docs.get_attributes('id'))
+        for idx in reversed(range(len(self._docs))):
+            if self._docs[idx].id in update_docs_ids:
+                update_doc = flat_docs.get_attributes_with_docs(self._docs[idx].id)[1][0]
+                del self._docs[idx]
+                self._docs.append(update_doc)
```
numb3r3, line 137:
```suggestion
        for doc in flat_docs:
            if doc.id is not None:
                self._docs[doc.id] = doc
            else:
                self._docs.append(doc)
        
```

numb3r3 (changes requested):

jinahub/indexers/SimpleIndexer/simple_indexer.py
alaeddine-13: can't you do:
```python
        delete_docs_ids = set(flat_docs.get_attributes('id'))
        for idx in delete_docs_ids:
            if idx in self._docs:
                del self._docs[idx]
```
?

jinahub/indexers/SimpleIndexer/simple_indexer.py
alaeddine-13: I do agree with @numb3r3 's suggestion.
simple tweak: maybe we should not accept updating docs with `None` ids or with ids that don't exists in the DAM

jinahub/indexers/SimpleIndexer/simple_indexer.py
numb3r3: oh, looks much better

Yongxuanzhang: what is the time complexity of `idx in self._docs` ? is it O(1)? If so then it is better. @alaeddine-13

numb3r3 (approved):
LGTM

PR #123, opened 2021-08-22
docs: add parameter details for custom image torch encoder

KING-SID:
Added parameter details to README for Custom Image Torch Encoder


JoanFM (approved):
Thank you very much for your contribution!

nan-wang (approved):
LGTM👍 However, in the long run, we'd better generate these directly from the docstrings.

PR #127, opened 2021-08-23
feat(faiss): load a trained index and add a training endpoint

numb3r3:
One additional **training endpoint** is introduced in this PR. 

To use a trainable faiss indexer (e.g., IVF based):

**1)** we can first apply a training flow to train a faiss indexer (absolutely, user can also implement a locally training script using native faiss api):
```python
from jina import Flow
import numpy as np

train_filepath = 'train.npy'
train_data = np.array(np.random.random([10240, 256]), dtype=np.float32)
np.save(train_filepath, train_data)

f = Flow().add(
    uses="jinahub://FaissSearcher",
    timeout_ready=-1,
    uses_with={
        'index_key': 'IVF10_HNSW32,PQ64',
        'trained_index_file': 'faiss.index',
        'on_gpu': False,
    },
)

with f:
    # the trained index will be dumped to "faiss.index"
    f.post(on='/train', parameters={'train_filepath': train_filepath})
```

**2)** Then in the query runtime, we can use FaissSearch by providing a pre-trained index file resulted from step 1, e.g., 
```python
f = Flow().add(
    uses="jinahub://FaissSearcher",
    timeout_ready=-1,
    uses_with={
        'index_key': 'IVF10_HNSW32,PQ64',
        'trained_index_file': 'faiss.index', # trained indexer 
        'on_gpu': False,
        'dump_path': '/path/to/dump_file'
    },
)
```

One limitation to mention is:
> Concurrent search/add or add/add are not supported.
https://github.com/facebookresearch/faiss/issues/367

Hence the `index` operation can not utilize multi-core CPUs to speed up.


JoanFM:
@numb3r3 concurrent operations should not be a problem, our parallelization model is based on multiprocess so one instance should have one index

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -41,29 +40,28 @@ class FaissSearcher(Executor):
...
-        # generate a training file in `.npy`
-        train_filepath = 'faiss_train'
-        np.save(train_filepath, train_data)
-        indexer = FaissIndexer('PCA64,FLAT', train_filepath)
     """
 
     def __init__(
         self,
         index_key: str = 'Flat',
-        train_filepath: Optional[str] = None,
```
JoanFM, line 59: so we do not support anymore training at initialization right? it seems a good change to me

numb3r3: yes, I created a separate training request endpoint to handle the training procedure.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -166,76 +162,96 @@ def _build_index(self, vecs_iter: Iterable['np.ndarray']):
...
+
+        index.nprobe = self.nprobe
+
+        self.index = self.to_device(index)
 
-        if self.requires_training:
-            if self.train_filepath:
-                train_data = self._load_training_data(self.train_filepath)
 
+    def _build_index(self, vecs_iter: Iterable['np.ndarray']):
```
maximilianwerk, line 179: I think `Iterable` is a too weak requirement here, since you iterate over `vecs_iter` several times.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -109,7 +108,11 @@ def __init__(
 
             self.num_dim = self._prefetch_data[0].shape[0]
             self.dtype = self._prefetch_data[0].dtype
-            self.index = self._build_index(vecs_iter)
+
+            if self.trained_index_file and not os.path.exists(self.trained_index_file):
+                raise ValueError(f'The trained index file {self.trained_index_file} does not exist')
+
+            self._build_index(vecs_iter)
```
maximilianwerk, line 115: That might be a stupid question due to my lack of understanding Faiss, but: Why do we need to build the index here? Shouldn't that be already part of the training? I would expect, that when I run the training method and restart the Searcher, it can directly pick up the trained AND filled index from disc. Or is this not possible/performant?

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -285,7 +301,74 @@ def search(
 
                 query_docs[doc_idx].matches.append(match)
 
-    def _train(self, index, data: 'np.ndarray', *args, **kwargs) -> None:
+    @requests(on='/train')
```
maximilianwerk, line 304: Do I get the following right:

We would have a TrainingFlow running all the time (or when training is needed), which is able to react an `train` requests. It will train a Faiss index, fill it with the latest data and dump it somewhere (potentially to S3 or so). Then a running FaissSearcher can be restarted with the new dump path and after loading the index directly start serving data.

I like it. My first reaction was "That is a miss-use of the Jina Flow concept!! How dare you!", but actually it has quite some charm. :)

maximilianwerk (changes requested):
Would it be possible to not only train, but also populate the index already during the `train` request?

Before merging, please wait for @cristianmtr feedback.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: Due to limited RAM, `Iterable` is chosen here to load large-scale data with a streaming method. To make the iteration safe, I iterate over the `vecs_iter` using batch mode (the batch_size is provided by the argument of `prefetch_size`)

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: > Why do we need to build the index here? Shouldn't that be already part of the training? 

1) with providing the `trained_index_file`, we can load indexer from `trained_index_file`, and then start to build the indexer.
2) without providing the `trained_index_file`, there are two options:
    2.1  train the faiss indexer, and then use method 1) to build the indexer;
    2.2 start the query flow, the searcher will train the indexer by using the indexing data (loading from dump datafile). 

> I would expect, that when I run the training method and restart the Searcher, it can directly pick up the trained AND filled index from disc. 

Yes,  the current PR is actually implementing what you described above.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -21,10 +22,8 @@ class FaissSearcher(Executor):
         Faiss package dependency is only required at the query time.
```
cristianmtr, line 22: Remove this

cristianmtr:
Can you update the README with some notes on this change?

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -151,94 +158,106 @@ def to_device(self, index, *args, **kwargs):
...
+                )
+                train_data = train_data[random_indices, :]
+
+            self.logger.info(f'Training Faiss indexer...')
+
+            if self.normalize:
+                faiss.normalize_L2(train_data)
+            self._train(train_data)
+
+        # TODO: Experimental features
```
cristianmtr, line 229: Do we want to keep this?

cristianmtr (changes requested):
Can you also add a full integration test that does

- start flow with empty index/ no training
- search -- no results
- train 
- search -- results!

And you can add it to the README as a showcase for how to use this feature.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: I think supporting training at initialization would be a good feature for some cases (e.g., small/media data). I would like to keep it.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
maximilianwerk: But you consume `vecs_iter` twice. Once in line 197 or 201 and once in the `_build_partial_index` call. You cannot consume an `Iterable` twice.

```def get_iter(count):
    for i in range(count):
        yield i


my_iter = get_iter(10)

first = list(my_iter)
second = list(my_iter)
print('first: ', first)
print('second: ', second)
```
outputs:
```
first:  [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
second:  []
```

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
maximilianwerk: I am pretty sure it isn't. The `_build_index` is still called on every startup. So let me rephrase this question on a different level:

Imagine you have a trained faiss model without data. And you have the same model filled with data and dump the filled index to disc (model + data is dumped).

Are the following two operations equally fast:

A. Loading the faiss model and populate it with data from a dump
B. just load the model + filled index from disc. No need to populate it with data, since it is already populated.

Is the B even possible?

If both are equally fast, we might still want to do B to not stress the PSQL database when to many shards ramp up.

Furthermore, doing B + incremental update would allow us to do performant query while indexing with short index-query delay.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: The elements fetching from the first iteration in line 197 are temporally filled into the list `self._prefetch_data`. And before the second iteration happened in the `_build_partial_index`, I first consume the data from `self._prefetch_data`. So, we did not lose data.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: ah,  you are right. I got misunderstanding your question. Yes, doing B is a great feature, and is planned to approach it in the next PR.

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
maximilianwerk: Ah I see. That makes a whole lot of sense. Thanks for the explanation.

cristianmtr (approved):

PR #130, opened 2021-08-24
feat: add key length to simple indexer

JoanFM:


numb3r3:
@JoanFM  this PR looks good to me. And I have a suggestion about the DAM implementation, is it reasonable to  throw a warning if the length of `doc.id` is larger than the `key_length`?

numb3r3 (approved):
LGTM

JoanFM:
> @JoanFM this PR looks good to me. And I have a suggestion about the DAM implementation, is it reasonable to throw a warning if the length of `doc.id` is larger than the `key_length`?

It would make sense, would u make a PR?

PR #138, opened 2021-08-27
Inverted index bm25

JoanFM:


nan-wang:
we decided not to implement this at the moment.

PR #146, opened 2021-08-29
ci: fix prep testbed

tadejsv:
Fixes pre-testbed script, so that it does not create an output matrix if no executors were changed


.github/workflows/ci.yml
```diff
@@ -47,6 +45,7 @@ jobs:
   unit-tests:
     needs: [prep-testbed]
     runs-on: ubuntu-latest
+    if: ${{ needs.prep-testbed.outputs.matrix != '[]' }}
```
cristianmtr, line 48: ah so if the matrix is empty the linters will run on the entire proj dir, right?

cristianmtr (approved):

PR #150, opened 2021-08-30
ci: notice changes in test folders

jakob1111996:
This branch fixes the bug that the CI does not detect changes in the test folders

### DIFF
--- a/scripts/get-all-test-paths.sh
+++ b/scripts/get-all-test-paths.sh
@@ -7,6 +7,14 @@ changed_folders=()
 
 for changed_file in $CHANGED_FILES; do
   file_base_dir=$(dirname $changed_file)
+  # Test folder changes
+  if [ $(basename $file_base_dir) = "tests" ]; then
+    file_base_dir=$(dirname "$file_base_dir")
+  fi
+  # Changes in subfolder of test folder (e.g. unit_test/integration)
+  if [ $(basename $(dirname "$file_base_dir")) = "tests" ]; then
+    file_base_dir=$(dirname $(dirname "$file_base_dir"))
+  fi
 
   # only if the folder has a tests or a Dockerfile but excluding integration tests (always run & separate)
   if [[ -f "${file_base_dir}/Dockerfile" || -d "${file_base_dir}/tests/" ]]; then


tadejsv:
I think we need to change the approach completely, and explicitly enumerate the allowed folders. Otherwise, there are too many cases to consider:
- what if it's a change in the `tests/integration`?
- what if it's a change in `executor/some_module/some_other_module`
- etc.

jakob1111996:
> 
> 
> I think we need to change the approach completely, and explicitly enumerate the allowed folders. Otherwise, there are too many cases to consider:
> 
>     * what if it's a change in the `tests/integration`?
> 
>     * what if it's a change in `executor/some_module/some_other_module`
> 
>     * etc.

How would you suggest we do that?

cristianmtr:
> I think we need to change the approach completely, and explicitly enumerate the allowed folders. Otherwise, there are too many cases to consider:
> 
> * what if it's a change in the `tests/integration`?

integration tests are always run

> * what if it's a change in `executor/some_module/some_other_module`

Good point. Maybe outside the scope of this PR, but I'll open an issue to track it

tadejsv:
> > I think we need to change the approach completely, and explicitly enumerate the allowed folders. Otherwise, there are too many cases to consider:
> > ```
> > * what if it's a change in the `tests/integration`?
> > 
> > * what if it's a change in `executor/some_module/some_other_module`
> > 
> > * etc.
> > ```
> 
> How would you suggest we do that?

My idea is something like this: list all the folders where executor directories can be, so this would be
- `jinahub/encoders/{text,audio,video}`
- `jinahub/rankers`
- `jinahub/segmenters`
- ...

There should only be a small number of them - and they should not change very much over time.

Here you list the **final** folder, right before the executor one. And then write a script that goes over these folders and collects the names of all the folders within them.

And then, in the part of the script that goes over changed files, check if any of these folders were matched.

Since this can be a pain to implement in bash, I would recommend doing it in python.

tadejsv:
> > I think we need to change the approach completely, and explicitly enumerate the allowed folders. Otherwise, there are too many cases to consider:
> > 
> > * what if it's a change in the `tests/integration`?
> 
> integration tests are always run
 
These are the executor integration tests

cristianmtr:
> > > I think we need to change the approach completely, and explicitly enumerate the allowed folders. Otherwise, there are too many cases to consider:
> > > ```
> > > * what if it's a change in the `tests/integration`?
> > > 
> > > * what if it's a change in `executor/some_module/some_other_module`
> > > 
> > > * etc.
> > > ```
> > 
> > 
> > How would you suggest we do that?
> 
> My idea is something like this: list all the folders where executor directories can be, so this would be
> 
> * `jinahub/encoders/{text,audio,video}`
> * `jinahub/rankers`
> * `jinahub/segmenters`
> * ...
> 
> There should only be a small number of them - and they should not change very much over time.
> 
> Here you list the **final** folder, right before the executor one. And then write a script that goes over these folders and collects the names of all the executor folders within them.
> 
> And then, in the part of the script that goes over changed files, check if any of these folders were matched.
> 
> Since this can be a pain to implement in bash, I would recommend doing it in python.

I think a simpler solution would be to keep going up until the the folder has a `tests` folder. That's what we care about in the end, right?

jakob1111996:
> I think a simpler solution would be to keep going up until the the folder has a `tests` folder. That's what we care about in the end, right?

This could be possible yes. We can just go up until there is a test folder -> then run CI or until we are in the root of the repository, then continue with the next changed file

cristianmtr (approved):

PR #155, opened 2021-08-31
fix: cd

tadejsv:
I noticed that due to the change in the CI script that gets test paths, the CD script broke. This fixes the problem

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -6,7 +6,7 @@ on:
       - main
   workflow_dispatch:
   # comment this out to test in CI
-  pull_request:
+  # pull_request:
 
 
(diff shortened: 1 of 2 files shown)


cristianmtr:
> I noticed that due to the change in the CI script that gets test paths, the CD script broke. This fixes the problem

Can you test it here? Uncomment this line and have the CD flow run here? https://github.com/jina-ai/executors/blob/2d00033d0e708b96477e1bd10bfcb76d597ffaed/.github/workflows/cd.yml#L9

You'll also need some mock changes in some executors.

tadejsv:
@cristianmtr Yep, it works

cristianmtr (approved):

PR #165, opened 2021-09-07
fix(faiss): casting from buffer

numb3r3:


numb3r3:
duplicate with https://github.com/jina-ai/executors/pull/166

PR #166, opened 2021-09-07
test: add integration test for sync with delta

cristianmtr:
Add an integration test to show off sync with deltas in PSQLFaissSearcher

### DIFF
--- a/jinahub/indexers/searcher/FaissSearcher/config.yml
+++ b/jinahub/indexers/searcher/FaissSearcher/config.yml
@@ -6,4 +6,6 @@ metas:
     - faiss_searcher.py
   name: faiss_searcher
   workspace: workspace/
+runtime_args:
+  name: faiss_searcher
 
--- a/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
+++ b/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
@@ -14,6 +14,10 @@ from jina.helper import batch_iterator
 from jina_commons import get_logger
 from jina_commons.indexers.dump import import_vectors
 
+GENERATOR_DELTA = Generator[
+    Tuple[str, Optional[np.ndarray], Optional[datetime]], None, None
+]
+
 
 class FaissSearcher(Executor):
     """Faiss-powered vector indexer
@@ -74,6 +78,7 @@ class FaissSearcher(Executor):
 
     def __init__(
         self,
+        # exhaustive search, which corresponds to what NumpySearcher was doing
         index_key: str = 'Flat',
         trained_index_file: Optional[str] = None,
         max_num_training_points: Optional[int] = None,
@@ -164,7 +169,11 @@ class FaissSearcher(Executor):
             self._doc_ids.append(id_)
             self._doc_id_to_offset[id_] = position
             self._is_deleted.append(0)
-            yield np.frombuffer(vector)
+            if vector is not None:
+                # this should already be a np.array, NOT bytes
+                yield vector
+            else:
+                yield None
 
     def device(self):
         """
@@ -585,7 +594,8 @@ class FaissSearcher(Executor):
             metric_type = faiss.METRIC_INNER_PRODUCT
         if self.metric not in {'inner_product', 'l2'}:
             self.logger.warning(
-                'Invalid distance metric for Faiss index construction. Defaulting to l2 distance'
+                'Invalid distance metric for Faiss index construction. Defaulting '
+                'to l2 distance'
             )
         return metric_type
 
@@ -600,29 +610,27 @@ class FaissSearcher(Executor):
             self._is_deleted.append(0)
         self._index(vecs)
 
-    def _add_delta(self, delta: Generator[Tuple[str, bytes, datetime], None, None]):
+    def _add_delta(self, delta: GENERATOR_DELTA):
         """
         Adding the delta data to the indexer
-        :param delta: a generator yielding (id, doc_vec_bytes, last_updated)
+        :param delta: a generator yielding (id, np.ndarray, last_updated)
         """
-        for doc_id, vec_buffer, _ in delta:
+        for doc_id, vec_array, _ in delta:
             idx = self._doc_id_to_offset.get(doc_id)
             if idx is None:  # add new item
-                if vec_buffer is None:
+                if vec_array is None:
                     continue
-                vec = np.frombuffer(vec_buffer, dtype=np.float32).reshape(
-                    1, -1
-                )  # shape [1, D]
+                # shape [1, D]
+                vec = vec_array.reshape(1, -1).astype(np.float32)
 
                 self._append_vecs_and_ids([doc_id], vec)
-            elif vec_buffer is None:  # soft delete
+            elif vec_array is None:  # soft delete
                 self._is_deleted[idx] = 1
             else:  # update
                 # first soft delete
                 self._is_deleted[idx] = 1
 
                 # then add the updated doc
-                vec = np.frombuffer(vec_buffer, dtype=np.float32).reshape(
-                    1, -1
-                )  # shape [1, D]
+                # shape [1, D]
+                vec = vec_array.reshape(1, -1).astype(np.float32)
                 self._append_vecs_and_ids([doc_id], vec)
--- a/jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
+++ b/jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
@@ -6,6 +6,7 @@ import datetime
 import functools
 from typing import Dict, Optional
 
+import numpy as np
 from jina import DocumentArray, Executor, requests
 from jina_commons import get_logger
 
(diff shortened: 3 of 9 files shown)


jinahub/indexers/searcher/FaissSearcher/config.yml
```diff
@@ -6,4 +6,6 @@ metas:
     - faiss_searcher.py
   name: faiss_searcher
   workspace: workspace/
+runtime_args:
+  name: faiss_searcher
```
numb3r3, line 10: Is it necessary? Where does `name` is used?

jinahub/indexers/searcher/compound/FaissPostgresSearcher/requirements.txt
```diff
@@ -2,3 +2,4 @@ psycopg2-binary==2.8.6
 git+https://github.com/jina-ai/executors
 faiss-cpu==1.6.5
 git+https://github.com/jina-ai/jina-commons
+git+https://github.com/jina-ai/jina.git@v2.0.23#egg=jina[hub]
```
numb3r3, line 5: Is it necessary?

jinahub/indexers/searcher/FaissSearcher/config.yml
cristianmtr: It's checked in a recent fix in master. But that fix required a subsequent fix, that wasn't merged at the time. Will try with latest core master

jinahub/indexers/searcher/compound/FaissPostgresSearcher/requirements.txt
cristianmtr: See https://github.com/jina-ai/executors/pull/166#discussion_r703636865

tests/integration/psql_import/test_import_psql.py
```diff
@@ -287,3 +287,89 @@ def test_start_up(docker_compose):
...
+    # for psql to start
+    time.sleep(2)
+
+    top_k = 5
+    batch_size = min(1000, nr_docs)
+    docs_original = get_batch_iterator(
+        batches=nr_docs // batch_size, batch_size=batch_size, emb_size=emb_size
+    )
+    docs_original = DocumentArray(flatten(docs_original))
+    assert len(docs_original) == nr_docs
```
tadejsv, line 312: Let's simplify - you only need `get_documents`, and no need to test jina core functionality

tests/integration/psql_import/test_import_psql.py
```diff
@@ -287,3 +287,89 @@ def test_start_up(docker_compose):
...
+
+    with _flow(uses_after=uses_after, total_shards=shards, startup_args={}) as flow:
+        flow.post(on='/index', inputs=docs_original)
+
+        results = flow.post(
+            on='/search',
+            inputs=get_documents(nr=1, index_start=0, emb_size=emb_size),
+            return_results=True,
+        )
+        assert len(results[0].docs[0].matches) == 0
```
tadejsv, line 326: This is just testing the `sync` call? Seems not to be what this test is about

tests/integration/psql_import/test_import_psql.py
```diff
@@ -287,3 +287,89 @@ def test_start_up(docker_compose):
...
+        )
+
+        results = flow.post(
+            on='/search',
+            inputs=get_documents(nr=3, index_start=0, emb_size=emb_size),
+            parameters={'top_k': top_k},
+            return_results=True,
+        )
+        assert len(results[0].docs[0].matches) == top_k
+        assert results[0].docs[0].matches[0].scores[METRIC].value > 0.0
```
tadejsv, line 343: Seems to me we are testing `search` here - this should be done (maybe already is) in a different test

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -610,9 +613,8 @@ def _add_delta(self, delta: Generator[Tuple[str, bytes, datetime], None, None]):
             if idx is None:  # add new item
                 if vec_buffer is None:
                     continue
-                vec = np.frombuffer(vec_buffer, dtype=np.float32).reshape(
-                    1, -1
-                )  # shape [1, D]
+                # shape [1, D]
+                vec = np.frombuffer(vec_buffer).reshape(1, -1).astype(np.float32)
```
maximilianwerk, line 617: I would prefer having PSQL directly store the right vectors data type and not do the type conversion here. Reasoning: This will shift time spend towards *indexing* time and not *syncing* time.

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
```diff
@@ -195,6 +195,12 @@ def update(self, docs: Optional[DocumentArray], parameters: Dict, **kwargs):
     @requests(on='/delete')
     def delete(self, docs: Optional[DocumentArray], parameters: Dict, **kwargs):
         """
-        Delete docs from PSQL, based on id
+        Delete docs from PSQL, based on id.
+
+        By default, it will be a soft delete, where the entry is left in the DB,
+        but its data will be set to None
         """
+        if 'soft_delete' not in parameters.keys():
```
maximilianwerk, line 203:
```suggestion
        if 'soft_delete' not in parameters:
```

maximilianwerk (dismissed):

tests/integration/psql_import/test_import_psql.py
cristianmtr: Can you elaborate?

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
cristianmtr: FaissSearcher could still receive this data from other sources. We still need it in there. I will make the default PSQL dtype be f32

jinahub/indexers/searcher/compound/FaissPostgresSearcher/faisspsql.py
cristianmtr: Done

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
maximilianwerk: Fair enough.

tests/integration/psql_import/test_import_psql.py
tadejsv: Sorry, what I mean is that this is not necessary. You are testing the `'use_delta': True` for the `sync` endpoint. Everything else can be assume to be working correctly in this test, and should be tested in a separate test if needed.

tests/integration/psql_import/test_import_psql.py
tadejsv: You are testing the case `'use_delta': True` in this test. Testing normal `sync` without it should be done as a separate test

tadejsv (changes requested):

cristianmtr:
@tadejsv I can't reply to your comments. 

The reason these tests are more complex 

1. They also act like showcases for how to full usage of these features.
2. `test_psql_import` needs to be the same as `test_dump_reload`, in order to compare them in the benchmark

> You are testing the case 'use_delta': True in this test. Testing normal sync without it should be done as a separate test

The use_delta only has meaning if you have already imported from a snapshot, since it will only give you the data from that last timestamp.

tadejsv:
Ok, so if I understand things correctly, the issue is that this unit test is trying to be 3 things at once: a unit test, a showcase, and a benchmark. Why not separate this out? 
- Showcase -> readme
-  benchmark -> perhaps the benchmarks repo? Not sure in what way you are using this benchmark

> The use_delta only has meaning if you have already imported from a snapshot, since it will only give you the data from that last timestamp.

My comment referred to the call to `/serach` endpoint after a normal sync, which is not really needed in that unit test. There can/should be another unit test testing normal sync.

P.S. For these comments you can reply under the original comments

cristianmtr:
> Ok, so if I understand things correctly, the issue is that this unit test is trying to be 3 things at once: a unit test, a showcase, and a benchmark. Why not separate this out?
> 
> * Showcase -> readme

It won't be tested and thus will end up broken.

> * benchmark -> perhaps the benchmarks repo? Not sure in what way you are using this benchmark

That would create too much coupling. We want it as both an integration test here, but, as you can see, it can also be called to run in `benchmark` mode with a larger nr of docs and larger emebddings.
  
> My comment referred to the call to `/serach` endpoint after a normal sync, which is not really needed in that unit test. There can/should be another unit test testing normal sync.

There is another unit test, in the PSQLFaissIndexer folder's `tests`. Yes, it can be removed. It's more like a gradual test, making sure that every step is ok before proceeding. So it can catch error early instead of waiting until the last assert.

tadejsv:
> That would create too much coupling. We want it as both an integration test here, but, as you can see, it can also be called to run in `benchmark` mode with a larger nr of docs and larger emebddings.

Well, fair point on coupling. But in general - a piece of code should do one thing and do that well. This tries to do too many things.

> There is another unit test, in the PSQLFaissIndexer folder's `tests`. Yes, the assert on the search could be removed. It's more like a gradual test, making sure that every step is ok before proceeding. So it can catch error early instead of waiting until the last assert.

Not just assert on the search, the whole search. There's no need to catch an error early - there should be other tests for that, so when something breaks, you just look at which tests fail and deduce the root cause from there. A test should focus on testing one thing - this makes them readable and easy to understand

cristianmtr:
> Not just assert on the search, the whole search. There's no need to catch an error early - there should be other tests for that, so when something breaks, you just look at which tests fail and deduce the root cause from there. A test should focus on testing one thing - this makes them readable and easy to understand

Ok, I'll remove that assert.

> Well, fair point on coupling. But in general - a piece of code should do one thing and do that well. This tries to do too many things.

Could be refactored later. Right now it would mean also changing the other benchmark tests (which would be a bit too much for this PR)

tadejsv (approved):
Seems good

PR #172, opened 2021-09-10
feat(gpu): split yolov5 cpu and gpu images

jacobowitz:
WIP - just opened to check sth on ci


jacobowitz:
To much work for now to fix the test, see this issue https://github.com/jina-ai/executors/issues/200

PR #174, opened 2021-09-13
docs: fix Simple README

cristianmtr:


tadejsv:
What is this bug?

cristianmtr:
It wasn't a bug. It had something to do with the workspace not being saved at container shutdown, when using it with `jinahub+docker`.

tadejsv:
Okay. And is there any fix that can be done for that? Since SimpleIndexer will probably be one of the most used executors, we should enable its use in a Docker container

cristianmtr:
Again, not sure it was an actual bug.

This works

```python
import numpy as np

from jina import Flow, Document

f = Flow().add(
    uses='jinahub+docker://SimpleIndexer',
    uses_with={'index_file_name': 'name'},
    uses_metas={'workspace': '/inner_workspace'},
    volumes=['workspace:/inner_workspace']
)

with f:
    resp = f.post(
        on='/index',
        inputs=[Document(id='a', embedding=np.array([1]))],
        return_results=True,
    )

with f:
    resp = f.post(
        on='/search',
        inputs=[Document(embedding=np.array([1]))],
        return_results=True,
        parameters={'top_k': 5},
    )
    assert resp[0].docs[0].matches[0].id == 'a'
```

tadejsv:
Okay, we should then also have a ticket to fix this issue

cristianmtr:
> Okay, we should then also have a ticket to fix this issue

What issue? It works. Might have been fixed in the meantime, if it was a problem

tadejsv:
Ah my bad, I thought you were adding the warning.

tadejsv (approved):

PR #177, opened 2021-09-13
fix(faiss): validate empty workspace folder

numb3r3:


jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
```diff
@@ -388,6 +388,10 @@ def save(self, target_path: Optional[str] = None, **kwargs):
         with open(os.path.join(target_path, 'delete_marks.bin'), "wb") as fp:
             pickle.dump(self._is_deleted, fp)
 
+    def _faiss_index_exist(self, folder_path: str):
+        index_path = os.path.join(folder_path, 'faiss.bin')
```
cristianmtr, line 392: Can you make `'faiss.bin'` a constant? Looks like it's hardcoded here and in the `save` method

cristianmtr (changes requested):
One comment

jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
numb3r3: addressed

cristianmtr (approved):

PR #178, opened 2021-09-13
fix: use python 3.8 in CI

alaeddine-13:
Executor normalizer needs python 3.8 (needs recent AST library)
closes https://github.com/jina-ai/executors/pull/178

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,9 +74,11 @@ jobs:
       - name: Python requirements
         run: pip install -r requirements.txt -r tests/requirements.txt
       - name: Create Dockerfile, if needed
-        run: |
-          if [[ ! -f "Dockerfile" ]]; then
-            docker run --entrypoint="normalizer" -v "$(pwd):/workspace" jinahub/hubble-normalizer:latest . --jina-version=2; sudo chown -R $(id -u ${USER}):$(id -g ${USER}) . ; fi
+          run: |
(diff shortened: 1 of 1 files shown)


tadejsv (changes requested):
We should use a container

alaeddine-13:
> We should use a container

can you re-check ?

tadejsv:
Ok then change looks good (you might consider setting `--user` for the container), and we need to then remove previous pip installation of the normalizer. Also, we need to test it somewhere (best place would be the normalizer repo)

cristianmtr:
For any CI/CD changes we need to test that all Executors still work.

1. Can you do a mock change in all the folders so that they are all triggered?
2. Can you enable the `jina hub push` job (see `cd.yml`) to run in the PR?

cristianmtr (dismissed):
See comment above

tadejsv:
@cristianmtr This is something that will take a lot of time, speaking from experience

cristianmtr:
> @cristianmtr This is something that will take a lot of time, speaking from experience

Fixing it afterwards also takes a lot of time.

tadejsv:
Let me create a quickfix then. The problem is that the progress on other PRs is blocked currently, because unit tests do not work due to new version of executor-normalizer

alaeddine-13:
dummy changes made and tested some of the executors
And then the changes are reverted

tadejsv (approved):
LGTM!

PR #188, opened 2021-09-14
fix: move executor constructor docstring under constructor

alaeddine-13:
fix: move executor constructor docstring under constructor


nan-wang (approved):
LGTM👍

PR #199, opened 2021-09-15
feat(gpu): split TorchObjectDetectionSegmenter in cpu and gpu

jacobowitz:
This is part of #169

It splits the TorchObjectDetectionSegmenter into cpu/gpu versions.
Also the previous device implementation was incomplete (not actually creating the model on GPU), this is fixed here


tadejsv (approved):
LGTM!

PR #222, opened 2021-09-20
chore: rename FaissPostgresIndexer

cristianmtr:


maximilianwerk (approved):
Has this some implication for hubble? If you don't know, can we first check before merging?

nan-wang:
> Has this some implication for hubble? 
Unfortunately, yes. They will be considered as new executors and therefore need new UUID and secrets

PR #226, opened 2021-09-21
refactor: refactor image normalizer

JoanFM:


jinahub/crafters/ImageNormalizer/normalizer.py
```diff
@@ -45,16 +44,16 @@ def __init__(
...
+        self.target_dtype = None
 
         # when passed from yaml it is string
         if isinstance(target_dtype, str):
             actual_type = locate(target_dtype)
             if actual_type:
                 self.target_dtype = actual_type
             else:
-                self.logger.error(
+                raise Exception(
```
bwanglzu, line 56: what type of exception do we want to captuer, maybe a bit more explicit?

bwanglzu (changes requested):
we be better if you create a normalizer as a fixture in test

jinahub/crafters/ImageNormalizer/normalizer.py
JoanFM: what exception would u raise?

JoanFM: Please review again

bwanglzu (approved):
LGTM!

PR #229, opened 2021-09-21
refactor: rankers

tadejsv:
- [x] DPRReaderRanker
- [x] SimpleRanker - I have decided to add this one, as it is actually used in our examples, and seems quite important. I have slightly generalized it, because the current structure of looking for matches in the chunks seemed very particular to our lyrics example - I see no reason why going to matches directly shouldn't be a part of this executor.


jinahub/rankers/DPRReaderRanker/README.md
```diff
@@ -2,8 +2,76 @@
 
  **DPRReaderRanker** performs a re-ranking of the matches for each document (question), as well as the answer spans extraction for each match. It uses the DPR Reader model to re-rank documents based on cross-attention between the question and the potential answer passages.
 
-The **DPR** model was originally proposed in [Dense Passage Retrieval for Open-Domain Question Answering](https://arxiv.org/abs/2004.04906).
+The **DPR** model was originally proposed in [Dense Passage Retrieval for Open-Domain Question Answering](https://arxiv.org/abs/2004.04906). This ranker can be used together with **[DPRTextEncoder](https://hub.jina.ai/executor/awl0jxog)** to use the entire
+DPR pipeline for QA tasks.
 
+## Usage
```
JoanFM, line 8: As I understand from latest comments, we do not want to have so much info in the README.

JoanFM (changes requested):

jinahub/rankers/DPRReaderRanker/README.md
tadejsv: I thought that refers only to things that could be "auto-generated". For the highlighted executors we were actually adding such more informative examples.

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -83,53 +80,46 @@ def rank(
...
+            `parameters={'traversal_paths': ['r'], 'batch_size': 10}`
         """
 
         if not docs:
             return None
 
-        for doc in docs.traverse_flat(
-            parameters.get('traversal_paths', self.traversal_paths)
+        for doc_arr in docs.batch(
+            batch_size=1,
```
bwanglzu, line 109: why batch size is hard coded here?

jinahub/rankers/SimpleRanker/simpleranker.py
```diff
@@ -1,69 +1,117 @@
...
-                    chunk_match_list.sort(key=lambda m: -m.scores[self.metric].value)
-                match = chunk_match_list[0]
-                match.id = chunk_match_list[0].parent_id
-                if self.ranking in ['mean_min', 'mean_max']:
-                    scores = [el.scores[self.metric].value for el in chunk_match_list]
-                    match.scores[self.metric] = sum(scores) / len(scores)
-                doc.matches.append(match)
+                    score = max(scores)
+                elif self.ranking in ['mean_min', 'mean_max']:
+                    score = sum(scores) / len(scores)
```
bwanglzu, line 107: do you think in edge cases we might get divide by 0 error? e.g. docs has value while no matches at `cm` level.

bwanglzu (changes requested):
minor comments

jinahub/rankers/SimpleRanker/simpleranker.py
```diff
@@ -1,69 +1,117 @@
...
+    document id, we add a final step where an indexer is used to retrieve the song
+    contents base on the id.
     """
 
     def __init__(
         self,
         metric: str = 'cosine',
         ranking: str = 'min',
         traversal_paths: Iterable[str] = ('r',),
+        matches_path: str = 'cm',
```
nan-wang, line 34: Do we need `travesal_paths` here? It is always from `chunks` to `matches`.

```suggestion
        matches_path: str = 'cm',
```

nan-wang: I see you point. We only need to keep either `traversal_paths` or `matches_path`. They are coupled because as long as one of them is fixed, the other is determined.

jinahub/rankers/SimpleRanker/simpleranker.py
```diff
@@ -1,69 +1,117 @@
...
         self.metric = metric
-        assert ranking in ['min', 'max', 'mean_min', 'mean_max']
         self.ranking = ranking
         self.traversal_paths = traversal_paths
+        self.matches_path = matches_path
 
     @requests(on='/search')
-    def rank(self, docs: DocumentArray, parameters: Dict, *args, **kwargs):
-        traversal_paths = parameters.get('traversal_paths', self.traversal_paths)
+    def rank(self, docs: DocumentArray, parameters: Dict, **kwargs):
```
nan-wang, line 75:
```suggestion
    def rank(self, docs: Optional[DocumentArray] = None, parameters: Dict = {}, **kwargs):
```

nan-wang (changes requested):

jinahub/rankers/DPRReaderRanker/README.md
```diff
@@ -2,8 +2,76 @@
...
+
+f = Flow().add(
+    uses='jinahub+docker://DPRReaderRanker', uses_with={'num_spans_per_match': 1}
+)
+
+
+def print_matches(response: Response):
+    for match in response.data.docs[0].matches:
+        score = match.scores['relevance_score'].value
+        print(f'Napoleon was born on {match.text} [relevance {score:.2%}]')
```
nan-wang, line 38: Add a comment and print `span_score` as well.

jinahub/rankers/DPRReaderRanker/dpr_reader.py
tadejsv: I see that this outer loop has caused some confusion. Let me explain what is happening here:
- first, we loop over documents in `docs` (doing this one by one)
- then, we loop over matches of a document in the outer loop (doing this in batches). This is the processing step where we actually use batching (to feed a batch of text to the model).

The outer loop can not really use batching, as we need to "attach" the new matches back to the original document. The use of `batch` there was just a convenience. I have now replaced it with `traverse_flat`, to avoid any misunderstanding. Please re-check @nan-wang @bwanglzu

jinahub/rankers/SimpleRanker/simpleranker.py
tadejsv: This is not possible. This stems from how `scores` are created - they are a value of `parents_scores` dict. That dict is created as a `defaultdict(list)`. So when I loop over matches, I append to this dictionary - this will create a list with scores for all `parent_id`s that appear. Conversely, for "non-existing" `parent_id`s no entry in dictionary will be created, thus there will be no empty lists to loop over.

I believe that this case (dividing by 0) should be covered by the `test_docs_no_matched_chunks` test. Let me know if you think we should test another case for this

jinahub/rankers/DPRReaderRanker/dpr_reader.py
```diff
@@ -83,53 +80,46 @@ def rank(
...
+            match_batches_generator = doc_arr.batch(
+                traversal_paths=['m'],
                 batch_size=parameters.get('batch_size', self.batch_size),
-                needs_attr='text',
+                require_attr='text',
             )
             for matches in match_batches_generator:
-                inputs = self._prepare_inputs(doc, matches)
-                with torch.no_grad():
+                inputs = self._prepare_inputs(doc_arr[0].text, matches)
```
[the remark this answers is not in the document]
tadejsv: Yes, everything works - see my comment on discussion above.

----

Here's the output for your example
```python
from jina import DocumentArray, Document
from dpr_reader import DPRReaderRanker

ranker = DPRReaderRanker()

docs = DocumentArray()
for i in range(2):
    q = Document(text=f'question {i}')
    for j in range(3):
        m = Document(text=f'question {i}: match {j}')
        q.matches.append(m)
    docs.append(q)

ranker.rank(docs=docs, parameters={})

for d in docs:
    print(f'matches for "{d.text}":')
    for i, m in enumerate(d.matches):
        print(f"Match {i}: {m.text}")
    print('*************************')
```
```console
matches for "question 0":
Match 0: question 0 : match 0 [SEP]
Match 1: question 0 : match 1 [SEP]
Match 2: question 0 : match 2 [SEP]
*************************
matches for "question 1":
Match 0: question 1 : match 1 [SEP]
Match 1: question 1 : match 0 [SEP]
Match 2: match 2 [SEP]
Match 3: question 1
*************************
```

jinahub/rankers/SimpleRanker/simpleranker.py
tadejsv: Well, not exactly. Here's the situation:

before my changes, this executor worked like this:
- first, it gathered all the documents that were in `traversal_paths` (call them "query documents")
- for the query documents, if gathered all matches of chunks (`matches_path="cm"`).

I thought that the second step was rather arbitrary and geared towards one of our applications - lyrics search, where the query text is broken into sentences as well, and then we use a searcher to add matches to these sentences (chunks). I saw no reason why you wouldn't want to "merge" matches directly under matches (`matches_path="m"`) of query documents. That's why I added this parameter.

Let me know if this is ok, or if it should be changed. I realize that this makes the executor more complex, but I think that without it it is just too "specific", especially for an executor named "SimpleRanker", which implies it is applicable to a wide range of basic use cases.

jinahub/rankers/DPRReaderRanker/README.md
tadejsv: Done

tadejsv:
@nan-wang @bwanglzu please re-check

bwanglzu (approved):
LGTM! Please confirm with hub channel that we want to keep the long or short version of readme.

jinahub/rankers/SimpleRanker/simpleranker.py
nan-wang: This is what I imaged about SimpleRanker
```
from jina import Document, DocumentArray
from jina.types.score import NamedScore

K = 3
M = 5

def loop(parent_doc: Document, k):
    chunk = Document(text=f'chunkk, g={k}', granularity=k+1)
    if k == K:
        for i in range(M):
            match = Document(text=f'match {i}')
            match.scores['cosine'] = NamedScore(value=i * 0.1)
            chunk.matches.append(match)
        parent_doc.chunks.append(chunk)
        return parent_doc
    else:
        loop(chunk, k+1)
        parent_doc.chunks.append(chunk)
        return parent_doc


def create_recursive_doc():
    return loop(Document(text='root'), 0)


if __name__ == '__main__':
    docs = DocumentArray([create_recursive_doc()])
    c = docs[0]
    for k in range(K):
        c = c.chunks[0]
    assert len(c.matches) == M

    ranker = SimpleRanker(matches_path='ccccm')
    ranker.rank(docs=docs)
    
    c = docs[0]
    for k in range(K-1):
        c = c.chunks[0]
    assert len(c.matches) == 1
    
```

nan-wang: image

nan-wang: image

tadejsv: @nan-wang Here is a modified version of your example that accomplishes what you want (according to the diagram you drew). Note the use of `traversal_paths` (which set for which document you want to create aggregated matches) and `matches_path` (which set how to get aggregated matches for that document). Let me know if this is ok


```python
from jina import Document, DocumentArray
from jina.types.score import NamedScore

from simpleranker import SimpleRanker

K = 3
M = 5


def loop(parent_doc: Document, k):
    chunk = Document(text=f'chunkk, g={k}', granularity=k + 1)
    if k == K:
        for i in range(M):
            match = Document(text=f'match {i}', parent_id='1')
            match.scores['cosine'] = i * 0.1
            chunk.matches.append(match)
        parent_doc.chunks.append(chunk)
        return parent_doc
    else:
        loop(chunk, k + 1)
        parent_doc.chunks.append(chunk)
        return parent_doc


def create_recursive_doc():
    return loop(Document(text='root'), 0)


if __name__ == '__main__':
    docs = DocumentArray([create_recursive_doc()])
    c = docs[0]
    for k in range(K + 1):
        c = c.chunks[0]

    assert c.granularity == K + 1  # 4
    assert len(c.matches) == M

    ranker = SimpleRanker(traversal_paths=['ccc'], matches_path='cm')
    ranker.rank(docs=docs)

    c = docs[0]
    for k in range(K):
        c = c.chunks[0]

    assert c.granularity == K  # 3
    assert len(c.matches) == 1
```

cristianmtr:
If this breaks the API, push the old version manually first, to have it. Like how we did for SimpleIndexer

jinahub/rankers/DPRReaderRanker/README.md
```diff
@@ -1,10 +1,89 @@
...
+```python
+f = Flow().add(
+    uses='jinahub+docker://DPRReaderRanker',
+    volumes='/your/home/dir/.cache/huggingface:/root/.cache/huggingface'
+)
+```
+
+### With GPU
+
+This encoder also offers a GPU version under the `gpu` tag. To use it, make sure to pass `device='cuda'`, as the initialization parameter, and `gpus='all'` when adding the containerized Executor to the Flow. See the [Executor on GPU](https://docs.jina.ai/tutorials/gpu-executor/) section of Jina documentation for more details.
```
cristianmtr, line 71: Just link to the docs

jinahub/rankers/SimpleRanker/simpleranker.py
```diff
@@ -1,69 +1,111 @@
...
+            want the ranker to work on - these are the "query" documents, for which
+            we wish to create aggregated matches.
+        :param matches_path: The path (starting from the "query" document) where the
+            matched chunks can be found. The default here is "cm", which means that
+            we expecte the matched chunks to be the matches of the chunks of our query
+            document. This path should end with "m"
         """
         super().__init__(*args, **kwargs)
+
+        if ranking not in ['min', 'max', 'mean_min', 'mean_max']:
```
cristianmtr, line 52: Make this a constant and use it here and in the `ValueError`

cristianmtr (changes requested):

jinahub/rankers/DPRReaderRanker/README.md
tadejsv: This is linking to the docs?

tadejsv:
@cristianmtr The changes are backwards compatible

jinahub/rankers/SimpleRanker/simpleranker.py
nan-wang: >         :param matches_path: The path (starting from the "query" document) where the
            matched chunks can be found.
I still think this is an overkill. This makes `SimpleRanker` not simple. 

> I think that without it it is just too "specific", especially for an executor named "SimpleRanker", which implies it is applicable to a wide range of basic use cases.

Being general for basic use cases is fine, but I think we are going too general and makes it over complicated

nan-wang:
```suggestion
```

jinahub/rankers/SimpleRanker/simpleranker.py
```diff
@@ -1,69 +1,111 @@
...
-            for chunk in doc.chunks:
-                matches_of_chunks.extend(chunk.matches)
-            groups = groupby(
-                sorted(matches_of_chunks, key=lambda d: d.parent_id),
-                lambda d: d.parent_id,
-            )
-            for key, group in groups:
-                chunk_match_list = list(group)
+            parents_scores = defaultdict(list)
+            for m in DocumentArray([doc]).traverse_flat([self.matches_path]):
```
nan-wang, line 89:
```suggestion
            for m in DocumentArray([doc]).traverse_flat(['m']):
```

nan-wang (changes requested):

jinahub/rankers/SimpleRanker/simpleranker.py
tadejsv: @nan-wang If we want to remove this argument, shouldn't we leave it at `'cm'` ? This was the original default case, and is used like this in the multrires-lyrics-search example.

nan-wang: Yes, you are right.

tadejsv:
@nan-wang ready to merge

nan-wang (approved):
LGTM👍

cristianmtr (approved):

PR #236, opened 2021-09-22
feat: improve TimmImageEncoder

tadejsv:


tadejsv:
Blocked until https://github.com/jina-ai/jina/pull/3451 gets released (due to batching on `blob`)

bwanglzu (approved):
I really like the use of `inference_mode`, maybe we could apply to other torch executors as well! LGTM!

PR #251, opened 2021-09-28
chore: remove jina commons

tadejsv:
Removes jina commons where it was used only for batching - this functionality is now superseded by `DocumentArray.batch`

### DIFF
--- a/jinahub/encoders/image/CLIPImageEncoder/clip_image.py
+++ b/jinahub/encoders/image/CLIPImageEncoder/clip_image.py
@@ -93,7 +93,7 @@ class CLIPImageEncoder(Executor):
 
         with torch.inference_mode():
             for batch_docs in document_batches_generator:
-                blob_batch = batch_docs.blobs
+                blob_batch = [d.blob for d in batch_docs]
                 if self.use_default_preprocessing:
                     tensor = self._generate_input_features(blob_batch.copy())
                 else:
@@ -107,7 +107,9 @@ class CLIPImageEncoder(Executor):
 
                 embeddings = self.model.get_image_features(**tensor)
                 embeddings = embeddings.cpu().numpy()
-                batch_docs.embeddings = embeddings
+
+                for doc, embed in zip(batch_docs, embeddings):
+                    doc.embedding = embed
 
     def _generate_input_features(self, images):
         input_tokens = self.preprocessor(
--- a/jinahub/encoders/image/CLIPImageEncoder/clip_image.py
+++ b/jinahub/encoders/image/CLIPImageEncoder/clip_image.py
@@ -3,7 +3,6 @@ from typing import Optional, Tuple
 import torch
 from jina import DocumentArray, Executor, requests
 from jina.logging.logger import JinaLogger
-from jina_commons.batching import get_docs_batch_generator
 from transformers import CLIPFeatureExtractor, CLIPModel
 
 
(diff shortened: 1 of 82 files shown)


jinahub/encoders/image/CLIPImageEncoder/clip_image.py
```diff
@@ -81,35 +80,36 @@ def encode(self, docs: Optional[DocumentArray], parameters: dict, **kwargs):
...
 
-                    embeddings = self.model.get_image_features(**tensor)
-                    embeddings = embeddings.cpu().numpy()
+                embeddings = self.model.get_image_features(**tensor)
+                embeddings = embeddings.cpu().numpy()
 
-                    for doc, embed in zip(batch_docs, embeddings):
-                        doc.embedding = embed
+                for doc, embed in zip(batch_docs, embeddings):
+                    doc.embedding = embed
```
bwanglzu, line 112: i see `batch` function create a generator of document arrays, so each of `batch_docs` should be a `DocumentArray`, then we can use the setter directly without looping? should be faster:

```python
da.embeddings = embeddings
```

it also applies to other loops

bwanglzu (changes requested):
Looks good to me, minor comment.

jinahub/encoders/image/CLIPImageEncoder/clip_image.py
tadejsv: Thanks, good observation

jinahub/encoders/image/CLIPImageEncoder/clip_image.py
```diff
@@ -81,35 +80,34 @@ def encode(self, docs: Optional[DocumentArray], parameters: dict, **kwargs):
...
+            require_attr="blob",
+        )
 
-                    embeddings = self.model.get_image_features(**tensor)
-                    embeddings = embeddings.cpu().numpy()
+        with torch.inference_mode():
+            for batch_docs in document_batches_generator:
+                blob_batch = batch_docs.blobs
+                if self.use_default_preprocessing:
+                    tensor = self._generate_input_features(blob_batch.copy())
```
JoanFM, line 98: why is the copy needed?

tadejsv: I thought it might otherwise modify the blob in place, not sure

jinahub/encoders/image/CLIPImageEncoder/clip_image.py
tadejsv: @bwanglzu it seems that this causes some issues - there might be something wrong with how we implement getters and setters in the core, I will investigate. For now I will revert my last commit, and try to add this feature back once I finish investigating the problem in another PR.

bwanglzu: sounds good to me!

bwanglzu (dismissed):
LGTM!

jinahub/encoders/image/CLIPImageEncoder/clip_image.py
JoanFM: I think it should not. the blob_batch is a numpy array with no access to the protobuf underlying

JoanFM: Also it is weird to change the input to generate something.

tadejsv: Okay, that makes sense

bwanglzu (approved):
LGTM!

PR #252, opened 2021-09-29
refactor: specify startup arg

cristianmtr:
Refactor usage of startup args, to be easier and adapt to scenario.

### DIFF
--- a/jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
+++ b/jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
@@ -90,5 +90,5 @@ Check [integration tests](https://github.com/jina-ai/executors/tree/main/tests/i
 
 For more advanced, full-fledged examples, check the [integration tests](https://github.com/jina-ai/executors/tree/main/tests/integration/psql_import).
 
-<!-- version=v0.1 -->
+<!-- version=v0.2 -->
 
--- a/jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
+++ b/jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
@@ -61,14 +61,26 @@ with get_my_flow() as flow:
     flow.post(
         on='/sync',
         # we need some specific parameters here
+        # startup is only needed the 1st time, to actually
+        # create the FaissSearcher
         parameters={
             'only_delta': True, 'startup': True
         }
     )
     flow.search(search_docs)
+    ...
+    flow.index(new_docs)
+    flow.post(
+        on='/sync',
+        # startup is not needed anymore
+        # FaissSearcher is incrementally updated
+        parameters={
+            'only_delta': True,
+        }
+    )
 ```
 
-While thi can **not** guarantee consistency, it is fast and has low overhead.
+While this can **not** guarantee consistency, it is fast and has low overhead.
 
 If you need consistency, you can use the **snapshot** method.
 This guarantees consistency between the shards, but at the cost of speed, extra disk usage, and an extra API call.
--- a/jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
+++ b/jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
@@ -57,27 +57,12 @@ Currently, we can only guarantee eventual consistency via manual delta updates.
 
 ```python
 with get_my_flow() as flow:
-    flow.index(your_docs)
-    flow.post(
-        on='/sync',
-        # we need some specific parameters here
-        # startup is only needed the 1st time, to actually
-        # create the FaissSearcher
-        parameters={
-            'only_delta': True, 'startup': True
-        }
-    )
-    flow.search(search_docs)
+    flow.post(on='/index', inputs=your_docs)
+    flow.post(on='/sync')
+    flow.post(on='/search', inputs=search_docs)
     ...
-    flow.index(new_docs)
-    flow.post(
-        on='/sync',
-        # startup is not needed anymore
-        # FaissSearcher is incrementally updated
-        parameters={
-            'only_delta': True,
-        }
-    )
+    flow.post(on='/index', inputs=new_docs)
+    flow.post(on='/sync')
 ```
 
 While this can **not** guarantee consistency, it is fast and has low overhead.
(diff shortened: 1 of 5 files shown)


jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
```diff
@@ -61,14 +61,24 @@ with get_my_flow() as flow:
     flow.post(
         on='/sync',
         # we need some specific parameters here
+        # startup is only needed the 1st time
```
maximilianwerk, line 64: Could you explain in one sentence why? Not sure if it helps the user, but it might.

cristianmtr: Re-check

jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
```diff
@@ -61,14 +61,26 @@ with get_my_flow() as flow:
     flow.post(
         on='/sync',
         # we need some specific parameters here
+        # startup is only needed the 1st time, to actually
+        # create the FaissSearcher
```
maximilianwerk, line 65: Sorry for being picky here, but from an APIs perspective these calls are strange. I know why we have the startup argument historically, but I believe we can simplify.

Let's look at the `_sync_only_delta` function (which we are talking about here essentially). We now use the very same function on `_kv_indexer` side. The only difference is, that we handle it differently to the `__init__` of Faiss and the `_add_delta` (which by the way should not be a private function, since we access it from the FaissPostgresIndexer). We could also always initialize a FaissSearcher when the FaissPostgresIndexer is initiallized and then only do `_add_delta` call. This would remove the `startup` parameter and make the user interface easier. Or do I miss some important point?

Since we want the delta update to be the default update, we can default `only_delta` to `True` and don't need any parameter on this call. That would be even more elegant :)

maximilianwerk: > We could also always initialize a FaissSearcher when the FaissPostgresIndexer is initiallized and then only do _add_delta call. 

Alternatively, we can check on each `_sync_only_delta` call, if a FaissSearcher exists. If not, create one.

numb3r3: >  remove the startup parameter and make the user interface easier.

I like this idea. Especially for the k8s deployment, it is impossible to determine whether it is the first time.

cristianmtr: Refactored. Check usage in `test_start_up`

maximilianwerk (commented):
I think I miss some use-case, but I still don't get, why we need `startup` at all. Couldn't we just do
```
        deltas = self._kv_indexer._get_delta(
            shard_id=self.runtime_args.pea_id,
            total_shards=self.total_shards,
            timestamp=timestamp,
        )
        # deltas will be like DOC_ID, OPERATION, DATA
        self._vec_indexer._add_delta(deltas)
```
in both cases inside `_sync_only_delta`? The `FaissSearcher` is already initialized, no reason to create a new one. Then `timestamp` can always default to `self._vec_indexer.last_timestamp`. This anyhow defaults to `datetime.min`.

The current semantic of `startup` is: Clear the FaissIndexer completely. If this is the use-case, we should rename the parameter to something like `recreate`. I'd argue, this should be done by recreate the full Pea for now.

cristianmtr:
> I think I miss some use-case, but I still don't get, why we need `startup` at all. Couldn't we just do
> 
> ```
>         deltas = self._kv_indexer._get_delta(
>             shard_id=self.runtime_args.pea_id,
>             total_shards=self.total_shards,
>             timestamp=timestamp,
>         )
>         # deltas will be like DOC_ID, OPERATION, DATA
>         self._vec_indexer._add_delta(deltas)
> ```
> 
> in both cases inside `_sync_only_delta`? The `FaissSearcher` is already initialized, no reason to create a new one. Then `timestamp` can always default to `self._vec_indexer.last_timestamp`. This anyhow defaults to `datetime.min`.
> 
> The current semantic of `startup` is: Clear the FaissIndexer completely. If this is the use-case, we should rename the parameter to something like `recreate`. I'd argue, this should be done by recreate the full Pea for now.

I don't think we can easily add delta if the index is empty. That's why it needs to recreate the FaissSearcher and call it with the constructor's `dump_func`

maximilianwerk:
> I don't think we can easily add delta if the index is empty. That's why it needs to recreate the FaissSearcher and call it with the constructor's dump_func

I see. So from Faiss perspective, the difference is: `startup = True` => Training must happen. This means, the very `sync` call defines the training data. When I get it right, whenever I set `startup=True`, the FaissSearcher will be retrained. Thus, I'd rename it to something like `train_faiss` or just `train`/`training`.

Apart from that, we should be very explicit in the Readme, that the first call to `sync` implies a train step. If you never call `sync` with `train=True` afterwards, the timing of the very first call heavily influences your search performance afterwards. This should be included as a `warning` or `note` or something in the Readme (if it not already is).

cristianmtr:
@maximilianwerk Renamed and adapted README

jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
```diff
@@ -96,8 +78,17 @@ All the others will fail and emit warning.
 To avoid this, you will need to split the Flow into two Flows, one for storage, and one for search, and use different polling methods.
 This is shown in the integration tests as well.
 
+### Via dump
+
+Alternatively, you can use the dump-reload method.
+
+See [docs](https://docs.jina.ai/advanced/experimental/indexers/#dump-and-rolling-update)
+
+Check [integration tests](https://github.com/jina-ai/executors/tree/main/tests/integration/psql_dump_reload) for an example on how to use it.
```
nan-wang, line 87: I suggest adding a README.md at https://github.com/jina-ai/executors/tree/main/tests/integration/psql_dump_reload. So that the uses have a better idea of how this works.

nan-wang (changes requested):

jinahub/indexers/searcher/compound/FaissPostgresIndexer/README.md
cristianmtr: Ok but that's outside the scope of this PR. I will add that in another PR. Any feedback on this one?

maximilianwerk (approved):
Really well done!

PR #256, opened 2021-10-01
refactor: hnswlib indexer

tadejsv:


jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -1,128 +1,271 @@
...
     def __init__(
         self,
-        default_top_k: int = 10,
+        top_k: int = 10,
         metric: str = 'cosine',
-        dump_path: Optional[str] = None,
-        default_traversal_paths: Optional[List[str]] = None,
-        is_distance: bool = False,
+        dim: int = 0,
+        max_elements: int = 1_000_000,
```
JoanFM, line 25: Is this needed? It feels like this should be Optional and in that case load everything there is

JoanFM (dismissed):

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: Yes, this is the parameter of the underlying library, `hnswlib`. It is needed.

jinahub/indexers/searcher/HnswlibSearcher/README.md
```diff
@@ -1,48 +1,92 @@
...
 
-- in the YAML definition
-  
-```yaml
-jtype: HnswlibSearcher
-with:
-    dump_path: /tmp/your_dump_location
-...
+    # Search for some docs
+    f.search(generate_index_docs(1000), show_progress=True)
```
cristianmtr, line 46: Rename it to just `generate_docs`

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -1,128 +1,271 @@
...
-        :param default_top_k: get tok k vectors
-        :param distance: distance can be 'l2', 'ip', or 'cosine'
-        :param dump_path: the path to load ids and vecs
-        :param traverse_path: traverse path on docs, e.g. ['r'], ['c']
-        :param reverse_score: True if add reversed distance as the `similarity` for match score, else return `distance` as score for match score.
-        :param ef_construction: defines a construction time/accuracy trade-off
-        :param ef_query:  sets the query time accuracy/speed trade-off
-        :param max_connection: defines tha maximum number of outgoing connections in the graph
+        :param top_k: Number of results to get for each query document in search
+        :param distance: Distance type, can be 'l2', 'ip', or 'cosine'
```
cristianmtr, line 36: metric or distance

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -1,128 +1,271 @@
...
+        )
+        self._ids_to_inds = bidict()
 
     @requests(on='/search')
-    def search(self, docs: Optional[DocumentArray], parameters: Dict, **kwargs):
+    def search(
+        self, docs: Optional[DocumentArray] = None, parameters: Dict = {}, **kwargs
+    ):
+        """Attach matches to the Documents in `docs`, each match containing only the
+        `id` of the matched document and the `score`.
```
cristianmtr, line 93: and the embedding, no?

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -1,128 +1,271 @@
...
+                f' the index ({self.dim})'
+            )
 
-        :return: Attaches matches to the Documents sent as inputs, with the id of the
-            match, and its embedding.
+        indices, dists = self._index.knn_query(docs_search.embeddings, k=top_k)
+
+        for i, (indices_i, dists_i) in enumerate(zip(indices, dists)):
+            for idx, dist in zip(indices_i, dists_i):
+                match = Document(id=self._ids_to_inds.inverse[idx])
```
cristianmtr, line 133: Should also add the embedding, or am I missing something?

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -1,128 +1,271 @@
...
+                match.scores[self.metric] = dist
+
+                docs_search[i].matches.append(match)
+
+    @requests(on='/index')
+    def index(
+        self, docs: Optional[DocumentArray] = None, parameters: Dict = {}, **kwargs
+    ):
+        """Index the Documents' embeddings. The documents should not be already
```
cristianmtr, line 142: What happens if I send a doc with the id of a previously indexed doc? This could happen quite easily and should be covered. Ignoring it, raising a warning etc.

jinahub/indexers/searcher/HnswlibSearcher/tests/integration/test_searcher.py
```diff
@@ -0,0 +1,90 @@
...
+
+        status_ind = f.post('/status', return_results=True)
+        status_ind = status_ind[0].data.docs[0].tags
+
+        assert status_ind['current_indexed'] == 2
+        assert status_ind['total_deleted'] == 0
+        assert status_ind['total_indexed'] == 2
+
+        result_search = f.search(da, return_results=True)
+        result_search = result_search[0].data.docs
```
cristianmtr, line 30: assert size of results. Else the `for` might be skipped if no results and the test will still pass.

jinahub/indexers/searcher/HnswlibSearcher/tests/integration/test_searcher.py
```diff
@@ -0,0 +1,90 @@
...
+        status_ind = f.post('/status', return_results=True)
+        status_ind = status_ind[0].data.docs[0].tags
+
+        assert status_ind['current_indexed'] == 2
+        assert status_ind['total_deleted'] == 0
+        assert status_ind['total_indexed'] == 2
+
+        # Check that we indeed have same items in index
+        result_search = f.search(da, return_results=True)
+        result_search = result_search[0].data.docs
```
cristianmtr, line 84: same

cristianmtr (changes requested):

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: No, see below

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: You are not missing anything - I have conciously decided against it.

Filling the document with attributes is the responsibility of the storage indexer. The search indexer is not even guaranteed to have the embedding (e.g. PQ Faiss), and may not be optimized for its retrieval even if it has it.

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: You are not missing anything - I have conciously decided against it.

Filling the document with attributes is the responsibility of the storage indexer. The search indexer is not even guaranteed to have the embedding (e.g. PQ Faiss), and may not be optimized for its retrieval even if it has it.

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: The issue I see here is that this involves checking that id is not already in the index for each item that we want to index. Even though this is executed in constant time (relative to size of index), if we index say 1 million documents, it could still add up (I calculated that it would add 10 seconds). If that's a delay we are willing tolerate, a check can be added

tadejsv: I was wrong, it would be only 0.1 seconds. Will change the logic so that items with existing id will get updated (easiest to implement and one that makes most sesnse imo)

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: Actually, distance is always returned, see Actually, distance is always returned, see https://github.com/nmslib/hnswlib#supported-distances

.github/workflows/ci.yml
```diff
@@ -84,7 +84,7 @@ jobs:
             jinahub/hubble-normalizer:v0.1.1 . --jina-version=2
             sudo chown -R $(id -u ${USER}):$(id -g ${USER}) . ; fi
       - name: Run unit tests
-        run: pytest -s -v -m "not gpu"
+        run: python -m pytest -s -v -m "not gpu"
```
nan-wang, line 87: What is the motivation for this?

jinahub/indexers/searcher/HnswlibSearcher/README.md
```diff
@@ -1,48 +1,92 @@
...
+            yield Document(embedding=np.random.rand(512))
 
-Check [tests](tests) for an example on how to use it.
+    return _gen_fn
 
-### Loading data
 
-Since this is a "Searcher"-type Executor, it does not _index_ new data. Rather they are write-once classes, which take as data source a `dump_path`. 
+f = Flow().add(
+    uses='jinahub+docker://HnswlibSearcher', uses_with={'dim': 512, 'top_k': 100}
```
nan-wang, line 34:
```suggestion
    uses='jinahub+docker://HnswlibSearcher/v0.2', uses_with={'dim': 512, 'top_k': 100}
```

jinahub/indexers/searcher/HnswlibSearcher/README.md
```diff
@@ -1,48 +1,92 @@
...
+    uses='jinahub+docker://HnswlibSearcher', 
+    uses_with={'dim': 512, 'top_k': 100, 'dump_path': '.'}
+)
+with f:
+    # Check that index was properly re-built
+    status = f.post('/status', return_results=True)
+    num_docs = int(status[0].data.docs[0].tags['current_indexed'])
+    print(f'{num_docs} documents in index')
+```
+
```
nan-wang, line 87: The descriptions for Deletion and updating operation are missing

nan-wang (changes requested):

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
cristianmtr: Can you make a note about that in the README? It will require the user to add another parameter to the search request in order to get the embedding from PSQL (see https://github.com/jina-ai/executors/blob/main/jinahub/indexers/storage/PostgreSQLStorage/postgres_indexer.py#L208)

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
cristianmtr: > items with existing id will get updated 

But a raise warning at the end with the list of ids that were updated.

.github/workflows/ci.yml
tadejsv: It's actually the proper way of adding the executor dir path to `sys.path` during testing (so that the executor will be imported). 

Currently what we have works only because we have `test/__init__.py` file, and this for some reason causes pytest to add the executor directory. In the future, pytest will not do that anymore (see `importlib` [here](https://docs.pytest.org/en/6.2.x/pythonpath.html#import-modes)).

In any case, this change has no effect on how things work now - as pytest [documentation](https://docs.pytest.org/en/6.2.x/usage.html#cmdline) explains, adding current dir to `sys.path` is pretty much the only difference.

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: Done

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: Currently it is already logging each such case (as info, not warning)

https://github.com/jina-ai/executors/blob/26cb8de69e514b6c83d7b5993063b125a588fa11/jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py#L172-L176

Do you think it needs to be a warning? I put it as info because I think it is most likely what the user actually intended (because overlapping ids have to be created explicitly)

jinahub/indexers/searcher/HnswlibSearcher/README.md
```diff
@@ -1,48 +1,146 @@
...
+This indexer has full support for CRUD operations, although only soft delete is possible.
 
+## Usage
 
+## Index and search
 
+This example shows a common usage pattern where we first index some documents, and then
+perform search on the index. 
 
+Note that to achieved the desired tradeoff between index and query
```
cristianmtr, line 16:
```suggestion
Note that to achieve the desired trade-off between index and query
```

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
cristianmtr: No, the name is different from the one in the constructor

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -1,128 +1,280 @@
...
-            if doc_idx is not None:
-                doc.embedding = np.array(self._indexer.get_items([int(doc_idx)])[0])
+
+        docs_to_update = docs.traverse_flat(traversal_paths)
+        if len(docs_to_update) == 0:
+            return
+
+        doc_inds, docs_filtered = [], []
+        for doc in docs_to_update:
+            if doc.id not in self._ids_to_inds:
```
cristianmtr, line 203: I would say to store these in one list and only emit one warning for all of them together. If you do one by one, it can slow down the process

cristianmtr (changes requested):

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: Ah ok, got it

tadejsv: Renamed all to `distance`

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
tadejsv: Printing all ids together would not look good in the output. Then I would just mention the number of documents that were updated

jinahub/indexers/searcher/HnswlibSearcher/README.md
```diff
@@ -1,48 +1,146 @@
...
+    uses='jinahub+docker://HnswlibSearcher/v0.2', uses_with={'dim': 512, 'top_k': 100}
+)
+with f:
+    # Index 10k docs
+    f.index(generate_docs(10_000), show_progress=True, request_size=100)
 
-This can be provided in different ways:
+    # Check that we have 10k docs in index
+    status = f.post('/status', return_results=True)
+    num_docs = int(status[0].data.docs[0].tags['current_indexed'])
```
nan-wang, line 47:
```suggestion
    num_docs = int(status[0].docs[0].tags['current_indexed'])
```

nan-wang (changes requested):

jinahub/indexers/searcher/HnswlibSearcher/README.md
```diff
@@ -1,48 +1,146 @@
...
+    uses='jinahub+docker://HnswlibSearcher/v0.2', uses_with={'dim': 512, 'top_k': 100}
+)
+with f:
+    # Index 10k docs
+    docs = list(generate_docs(10_000))
+    f.index(docs, show_progress=True, request_size=100)
+
+    # Delete the last 100 docs
+    ids_delete = [doc.id for doc in docs[-100:]]
+    f.post('/delete', parameters={'ids': ids_delete})
```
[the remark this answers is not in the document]
tadejsv: Done

jinahub/indexers/searcher/HnswlibSearcher/tests/integration/test_searcher.py
```diff
@@ -0,0 +1,92 @@
...
+            Document(id='a', embedding=np.ones(_DIM) * 1.0),
+            Document(id='b', embedding=np.ones(_DIM) * 2.0),
+        ]
+    )
+
+    with f:
+        f.index(da)
+
+        status_ind = f.post('/status', return_results=True)
+        status_ind = status_ind[0].data.docs[0].tags
```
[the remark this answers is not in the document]
tadejsv: Done

tadejsv:
@nan-wang please re-check

PR #257, opened 2021-10-01
feat: add video loader

makram93:
Video Loader that loads frames and audio as chunks which can be retrieved by using `modality='audio' or modality='image'`.


tadejsv:
Two comments:
- if you want to create temporary files/directories, use python's `tempfile` library
- you might not even need that. You can use ffmpeg to read the video directly to frames, see https://github.com/kkroening/ffmpeg-python/blob/master/examples/README.md#convert-video-to-numpy-array . For audio it might be a bit harder, but I think still possible?

jinahub/loaders/VideoLoader/README.md
```diff
@@ -0,0 +1,23 @@
...
+#### via source code
+
+```python
+from jina import Flow
+	
+f = Flow().add(uses='jinahub://VideoLoader')
+```
+
+- To override `__init__` args & kwargs, use `.add(..., uses_with: {'key': 'value'})`
+- To override class metas, use `.add(..., uses_metas: {'key': 'value})`
```
nan-wang, line 23: delete. Check out https://docs.google.com/document/d/1UebHh5WDArscXZkT5TKENDdsce0n9KmvkHJkoucBT_w/edit

nan-wang (changes requested):

CatStark (changes requested):
Can you move this outside from this repo please? It should be on its own repo

PR #265, opened 2021-10-13
fix: pass pea_id runtime arg

numb3r3:


jinahub/indexers/storage/PostgreSQLStorage/postgreshandler.py
```diff
@@ -128,6 +128,8 @@ def _create_table(self):
                 shard int,
                 last_updated timestamp with time zone default current_timestamp
             );
+            CREATE INDEX idx_shard ON {self.table}(shard);
```
cristianmtr, line 131: I think we talked about this and decided not to add an index as it would slow down? @maximilianwerk

jinahub/indexers/compound/FaissPostgresIndexer/faisspsql.py
```diff
@@ -73,7 +73,12 @@ def __init__(
     def _init_executors(self, dump_path, kwargs, startup_sync_args):
         # float32 because that's what faiss expects
         kv_indexer = PostgreSQLStorage(dump_dtype=np.float32, **kwargs)
-        vec_indexer = FaissSearcher(dump_path=dump_path, prefetch_size=16, **kwargs)
+        vec_indexer = FaissSearcher(
+            dump_path=dump_path,
+            prefetch_size=16,
+            shard_id=self.runtime_args.pea_id,
```
cristianmtr, line 79: I'm confused about how this could happen. The **kwargs includes `runtime_args.pea_id`

numb3r3: ah, I got a misunderstanding of `runtime_args`. This revision is not necessary.

PR #266, opened 2021-10-13
fix: missing model files in audioclip docker containers

jakob1111996:
When running the docker container versions of the AudioCLIP executors (AudioCLIPImageEncoder, AudioCLIPTextEncoder), the docker containers don't work out of the box because the model files are not downloaded during the container creation.

I would say that this is not desired behavior. In my opinion, the containers should work out-of-the-box. If I want to download the models myself, I wouse the jinabox://AudioCLIPImageEncoder syntax and not jinahub+docker://...

To fix this, I have added a flag `download_model` that specifies whether to download a model in the constructor of the executor.
This way, it is possible to use a model in a volume via download_model = False and alternatively have the model downloaded automatically.


cristianmtr:
Don't all encoders encourage the pattern of mounting volumes?

PR #286, opened 2021-10-27
feat: support removing fields from response

Ahmad-Nouh:
support removing fields from FaissLMDBSearcher response

### DIFF
--- a/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
+++ b/jinahub/indexers/searcher/FaissSearcher/faiss_searcher.py
@@ -49,22 +49,22 @@ class FaissSearcher(Executor):
     """
 
     def __init__(
-        self,
-        # exhaustive search, which corresponds to what NumpySearcher was doing
-        index_key: str = 'Flat',
-        trained_index_file: Optional[str] = None,
-        max_num_training_points: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


cristianmtr:
Hey @Ahmad-Nouh 

Would you have time to polish this for merging?

Ahmad-Nouh:
> Hey @Ahmad-Nouh
> 
> Would you have time to polish this for merging?

Hey @cristianmtr 
I couldn't find a way to implement this feature because it seems that the final response fields are being added from somewhere else outside the FaissLMDBSearcher.
I also tried to create an executor on top of the FaissLMDBSearcher to slim the response.
Although the executor output was free of extra fields, the extra fields were present in the last output,
It seems that they are being added from somewhere else.

cristianmtr:
> > Hey @Ahmad-Nouh
> > Would you have time to polish this for merging?
> 
> Hey @cristianmtr I couldn't find a way to implement this feature because it seems that the final response fields are being added from somewhere else outside the FaissLMDBSearcher. I also tried to create an executor on top of the FaissLMDBSearcher to slim the response. Although the executor output was free of extra fields, the extra fields were present in the last output, It seems that they are being added from somewhere else.

I see. Do you have some code where this happens? I would expect that removing those fields would work.

PR #289, opened 2021-11-03
fix: add filter in hnsw searcher

numb3r3:


jinahub/indexers/searcher/HnswlibSearcher/tests/unit/test_searcher.py
```diff
@@ -80,6 +80,12 @@ def test_index():
     assert index._index.element_count == NUM_DOCS
     assert set(index._ids_to_inds.keys()) == set(da1.get_attributes('id'))
 
+    da3 = DocumentArray([Document() for emb in embeddings])
```
tadejsv, line 83: Can you move this out to a separate test function? I prefer having each unit test as atomic as possible

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -160,7 +160,9 @@ def index(
         if docs is None:
             return
 
-        docs_to_update = docs.traverse_flat(traversal_paths)
+        docs_to_update = docs.traverse_flat(
```
cristianmtr, line 163: As agreed, make this a parameter to control whether or not this filter is applied

jinahub/indexers/searcher/HnswlibSearcher/tests/unit/test_searcher.py
numb3r3: addressed in latest commits

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         dump_path: Optional[str] = None,
         traversal_paths: Iterable[str] = ('r',),
         is_distance: bool = True,
+        ignore_invalid_docs: bool = True,
```
nan-wang, line 32:
```suggestion
        check_embedding: bool = True,
```

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -46,6 +47,7 @@ def __init__(
         :param traversal_paths: The default traverseal path on docs (used for indexing,
             search and update), e.g. ['r'], ['c']
         :param is_distance: Boolean flag that describes if distance metric need to be reinterpreted as similarities.
+        :param ignore_invalid_docs: Boolean flat that determine whether filter out invalid docs (i.e., without embedding)
```
nan-wang, line 50:
```suggestion
        :param check_embedding: If `check_embedding` is False (default: True), then the validation of embedding   will be skipped. Documents without embeddings will result in an `TypeError` or `ValueError`.
```

nan-wang (changes requested):

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
numb3r3: the parameter name`check_embedding=True` cannot indicate the action of filtering.  Could the users confuse this parameter?

cristianmtr: I'd propose `filter_invalid_docs`

jinahub/indexers/searcher/HnswlibSearcher/hnswlib_searcher.py
```diff
@@ -164,7 +167,19 @@ def index(
         if len(docs_to_update) == 0:
             return
 
-        embeddings = docs_to_update.embeddings
+        try:
+            embeddings = docs_to_update.embeddings
+        except (TypeError, ValueError) as ex:
+            if self.ignore_invalid_docs:
+                docs_to_update = docs_to_update.traverse_flat(
+                    ['r'], filter_fn=lambda d: d.embedding is not None
```
cristianmtr, line 175: and `d.embedding.shape == self.blablabla.shape`, no?

numb3r3: I would say the mis-matched shape can directly raise an exception.

cristianmtr: Yesterday we agreed that Indexer should not implement any filtering. It is the job of the upstream to make sure documents are correct. So this PR should be closed, right?

numb3r3: yes, we can close this PR.

numb3r3:
The filter should be done in upstream.

PR #290, opened 2021-11-04
Fix faiss bad delta recover

cristianmtr:
- only add ids if the delta update was successful and test to show this


cristianmtr:
Real PR https://github.com/jina-ai/executors/pull/291

PR #291, opened 2021-11-04
fix: add ids only if good delta op

cristianmtr:
- only add ids if the delta update was successful and test to show this


nan-wang (approved):
LGTM👍

nan-wang (approved):
LGTM👍

PR #296, opened 2021-11-08
fix: add tests for multi shards

nan-wang:
Add multi-shard support to `HnswlibSearcher`


nan-wang:
this is only for demonstration purpose.

PR #299, opened 2021-11-16
fix: cleanup endpoint correct path

cristianmtr:


maximilianwerk (dismissed):
LGTM. Feels like a noob question, but do wee need to bump some version somewhere?

cristianmtr:
> LGTM. Feels like a noob question, but do wee need to bump some version somewhere?

not a noob question. I was the noob for forgetting to do it :P

maximilianwerk (approved):
LGTM

